PR #27, opened 2026-04-23
feat(clients): organizational unit placement — PR 1 (intake + history display)

lars-tice:
## Summary

First of a three-PR series landing the `client-ou-edit` feature. PR 1 introduces organizational-unit (OU) tracking on client placements — intake form picker, placement-history display, backend schema/RPCs, and the `client.transfer` permission.

Subsequent PRs: **PR 2a** (client edit mode — Demographics + Admission), **PR 2b** (remaining sections + sub-entity CRUD). Phase 9 (activate the parked `api-rpc-readback-pattern` follow-up) triggers immediately after this PR merges.

### Backend (Supabase)
- **Migration `20260422052825`** — `client_placement_history_projection.organization_unit_id` (nullable FK, partial index on `WHERE is_current = true`); `api.change_client_placement()` extended with `p_organization_unit_id`; `handle_client_placement_changed()` now extracts OU, takes a `FOR UPDATE` lock on the prior current row, and denormalizes OU + arrangement to `clients_projection`; `handle_client_information_updated()` and `handle_client_admitted()` no longer write `organization_unit_id` (single-path mutation); `client.transfer` permission seeded with role-template + implication edges + backfill for active `provider_admin` roles; `api.update_client()` picks up the read-back + processing_error pattern (narrow-scope proof-of-pattern; full rollout is the parked Phase 9 follow-up).
- **Migration `20260423013804`** (Phase 6 read-model enrichment) — `api.get_client()` LEFT JOIN surfaces `organization_unit_name`, `organization_unit_is_active`, `organization_unit_deleted_at` on each placement_history item. Architect-reviewed ([software-architect-dbc](documentation/architecture/decisions/adr-client-ou-placement.md#decision-5--ou-current-state-enriched-at-read-time-not-on-the-history-row-phase-6)) over two alternatives (denormalize onto projection / filter the JOIN) both of which would violate event-sourced audit semantics.

### Frontend
- `ClientPlacementHistory` type extended with OU id + read-time state fields; `ChangePlacementParams` accepts `organization_unit_id?`.
- `ClientIntakeFormViewModel` loads OUs via `IOrganizationUnitService`, renders a `TreeSelectDropdown` on the Admission section (`data-testid=\"admission-ou-select\"`), and emits `change_client_placement` when all three (arrangement + OU + admission_date) are set.
- `PlacementCard` on `ClientOverviewPage` renders a three-state OU label (name / `name (inactive)` / `—`) via `formatPlacementOuLabel()` helper wrapped in `data-testid=\"placement-ou-label\"`.
- `organizationUnitPath.ts` utility + unit tests for id↔ltree-path bridging between TreeSelectDropdown and VM formData.
- `MockClientService.getClient()` resolves current OU state per placement row via the `MockOrganizationUnitService` singleton — mirrors the real `api.get_client()` LEFT JOIN so mock/production E2E exercise the same rendering rule.

### Documentation
- **New ADR**: [adr-client-ou-placement.md](documentation/architecture/decisions/adr-client-ou-placement.md) — 5 decisions (single-path OU mutation, `client.transfer` permission, row lock, arrangement fallback, read-time OU state enrichment).
- Table docs updated: `client_placement_history_projection.md` (new column + FK + partial index + mutation path), `clients_projection.md` (denormalized-but-not-directly-mutable note).
- `AGENT-INDEX.md` extends `client-placement`, adds `client-transfer` keyword, catalogs the new ADR.

### Bundled housekeeping (scope-independent)
Three commits repair unrelated SKILL.md/AGENT-INDEX drift: `effc0dd9` (frontend-dev-guidelines), `53cf2fcb` (three ViewModel-family docs cataloged), `91a6c467` (infrastructure-guidelines edge-function scope). These were authored separately and landed on this branch mid-feature; reviewing alongside is fine but they're not gated on the feature.

## Architectural decisions (see the ADR for full rationale)

1. **Single-path OU mutation** — only `client.placement.changed` writes `organization_unit_id`; prior side-effect writes on `client.information.updated` / `client.admitted` removed.
2. **`client.transfer` ≠ `client.update`** — transfers have different audit requirements than chart edits; clinicians should not inherit transfer rights. Intake OU picker is gated on `client.create` (initial placement is not a transfer).
3. **Row lock in placement handler** — `FOR UPDATE` on prior `is_current = true` row serializes concurrent placement changes at the database level.
4. **Arrangement fallback on OU-only edit** — edit ViewModel reuses `originalFormData.placement_arrangement` when the user changes OU but not arrangement (Phase 5a).
5. **OU state at read time, not on the history row** — `LEFT JOIN` in `api.get_client()` annotates current state; history rows stay immutable per event-sourcing audit semantics.

## Test plan

- [x] `npm run typecheck` (frontend) — 0 errors
- [x] `npm run lint` (frontend) — 0 warnings
- [x] `npm run docs:check` (frontend) — 0 issues, 100% coverage
- [x] `npm run test -- --run src/utils src/viewModels/client src/pages/clients src/services/clients` — 67 passed
- [x] `npx playwright test --config playwright.client-intake.config.ts` — 58 passed (includes 4 new OU-placement tests: picker renders with seed OUs, inactive units filtered out, no-OU intake → no placement section, OU intake → placement card shows OU name without \"(inactive)\")
- [x] `supabase db push --linked --dry-run` — clean
- [x] Migrations applied to linked project; MCP SQL verification 12/12 structural assertions pass (column/FK/index, RPC signature, permission seed + template, handler OU-removal assertions, placement handler `FOR UPDATE` lock, Phase 6 get_client response keys, backfill sanity)
- [x] Failed-events check: 0 `client.placement.*` events with processing_error after migration apply
- [ ] **Reviewer to verify**: manual smoke in mock mode — intake with OU selection ends with a placement history card showing \"Main Campus Medical Center\"; intake without OU selection results in no placement section on detail page
- [ ] **Reviewer to verify**: backfill sanity in staging/prod (`SELECT COUNT(*) FROM client_placement_history_projection WHERE is_current=true AND organization_unit_id IS NULL`) — acceptable for orgs with clients not yet assigned to OUs

## Notes for reviewer

- **Migration ordering**: two migrations in this PR. `20260422052825` is the substantive schema + handler + RPC + permission seed work (applied 2026-04-22). `20260423013804` is the additive Phase 6 read-model enrichment (applied 2026-04-23). Both are idempotent `CREATE OR REPLACE` / guarded DDL.
- **Handler reference files** (Rule 7b): 3 updated to match live DB — `handle_client_placement_changed.sql`, `handle_client_information_updated.sql`, `handle_client_admitted.sql`. `api.*` RPCs are not tracked under `infrastructure/supabase/handlers/` by convention.
- **Follow-up (Phase 9)**: after this PR merges, the parked `dev/parked/api-rpc-readback-pattern/` feature must be activated (move to `dev/active/`, start PR 1 of that feature). `api.update_client` in this PR is the only instance of the read-back + processing_error pattern so far; the rest of the codebase still has silent-failure surface area. This feature is **not** archivable without Phase 9 executed.

## Dev-docs trail

- Feature tasks/context/plan: [`dev/active/client-ou-edit-*.md`](dev/active/)
- Parked follow-up: [`dev/parked/api-rpc-readback-pattern/`](dev/parked/api-rpc-readback-pattern/)


lars-tice:
## Code Review

### Overview

PR 1 of a 3-PR series adds OU tracking to client placements. Delivers a coherent vertical slice: AsyncAPI contract → migration (schema + RPCs + handlers + permission seed + backfill) → frontend types/services → intake UI → read-time display → E2E coverage → ADR + table docs. Also bundles three unrelated SKILL.md/AGENT-INDEX drift repairs (explicitly called out in PR body).

**Scale**: +3,186 / −258 across 32 files. Bundled skill/docs housekeeping accounts for ~600 lines; feature proper is well-scoped.

### Strengths

- **Architect-reviewed**: Visible `software-architect-dbc` integration with labeled decisions (C3/C4/M1/M5/M7/M8). ADR captures 5 load-bearing decisions cleanly, including alternatives that were rejected (Decision 5).
- **Single-path OU mutation** (C3) is enforced by *removing* the dual-write branches from `handle_client_information_updated` and `handle_client_admitted` — not just by policy. Correct way to prevent projection drift.
- **Concurrency**: `FOR UPDATE` lock in `handle_client_placement_changed` is correct. Two racing events serialize on the prior `is_current=true` row; READ COMMITTED re-evaluates the WHERE clause post-lock-wait so T2 sees T1's new row.
- **Idempotent migration**: `IF NOT EXISTS`, `ON CONFLICT DO NOTHING`, `IS NULL` backfill guard, and explicit `DROP FUNCTION ... (old signature)` before adding the new overload (avoids the stale-overload footgun the context notes).
- **Contract-first discipline**: AsyncAPI updated before migration; bundle + generated types regenerated.
- **Read-time OU state enrichment** (Phase 6) is architecturally principled — history rows stay immutable, current OU state comes from JOIN. Both alternatives were considered and rejected with clear rationale.
- **Mock parity**: `MockClientService.getClient` resolves OU state per placement row, matching the real `api.get_client` LEFT JOIN, so E2E exercises the same rendering path.
- **Handler reference files updated** (Rule 7b) for all three modified handlers.
- **RLS unchanged and explicitly documented** in migration header (M7).

### Issues to Address

#### 🔴 Major: Permission check mismatch — \`client.transfer\` is seeded but never enforced

\`api.change_client_placement\` (migration 1c) checks:

\`\`\`sql
IF NOT public.has_effective_permission('client.update', v_org_path) THEN
\`\`\`

But ADR Decision 2 introduces \`client.transfer\` precisely so clinicians (who have \`client.update\` for chart edits) **don't** inherit transfer rights. As written, any user with \`client.update\` can call this RPC — \`client.transfer\` is seeded, implications are wired, provider_admin gets backfilled, but no code path ever *checks* it.

The implication graph goes the wrong way for enforcement: \`client.transfer → client.update\` means a transferrer has update (good for role design), but an updater does **not** have transfer — yet the RPC only gates on update. This defeats the whole M1 decision.

Neither the PR body, the ADR, nor the migration header acknowledge this gap. The ADR reads as if the permission model is in force; it isn't.

**Recommendation**: Either

- (a) Have the RPC accept a \`p_is_transfer boolean\` (or infer from context: \"transfer = client already has an is_current placement\") and branch on \`client.transfer\` vs \`client.create\`, or
- (b) Defer to PR 2a and explicitly document in the ADR + PR body that \`client.transfer\` is seeded-but-unused at the enforcement layer until then — and that a user with only \`client.update\` can currently call the RPC to effect a transfer.

Either way, this needs to be called out explicitly before merge.

#### 🟡 Minor: Potential same-day placement UNIQUE violation

\`client_placement_history_projection\` has \`UNIQUE (client_id, start_date)\`. \`handle_client_placement_changed\`'s \`ON CONFLICT\` clause handles only \`client_placement_history_projection_pkey\` — not the \`(client_id, start_date)\` constraint. Two placement changes for the same client on the same day (rare but possible: admin corrects an OU selection minutes after initial intake) would surface as an unhandled constraint violation → \`processing_error\` on the event.

Not critical for PR 1 (unlikely in practice), but worth either handling in \`ON CONFLICT\` or noting as accepted risk in the ADR.

#### 🟡 Minor: Test scope drift

\`'registering a client without selecting an OU shows no placement history'\` also leaves \`placement_arrangement\` empty, so it's actually testing \"no placement event emitted when any of the three required fields is missing\" rather than the OU-specific scenario. The inline comment acknowledges this, but the test name is misleading. Consider splitting into two tests or renaming to reflect the actual guard being exercised.

#### 🟡 Minor: \`Partial<Client>\` widening loses type safety

\`ClientRpcResult.client\` widened to \`Partial<Client>\` — correct given the projection read-back doesn't populate sub-entity arrays, but consumers now get no compile-time hint about which fields are actually present. Consider a separate \`ClientProjectionRow\` type that mirrors exactly what the RPC returns, rather than \"any subset of Client might or might not be there.\"

#### 🟢 Nit: Bundled scope

The three SKILL.md/AGENT-INDEX repairs (commits \`effc0dd9\`, \`53cf2fcb\`, \`91a6c467\`, \`f52aa35c\`) are substantial content drift fixes, not mechanical cleanup. PR body flags them honestly. They're valuable and well-researched (architect-reviewed), but they substantially enlarge the review surface. Future consideration: land housekeeping PRs separately — easier to revert either piece without affecting the other.

### Test Coverage

- ✅ 58 Playwright intake tests (4 new OU-specific) pass
- ✅ 67 Vitest tests across utils/viewModels/pages/services pass
- ✅ 12/12 structural SQL assertions pass against linked project
- ✅ Unit tests for \`organizationUnitPath\` utility cover null/empty/prefix-non-match/round-trip
- ⚠️ Functional SQL test (emit event → assert projection state) deferred because dev DB has 0 clients — acceptable given the handler body is structurally asserted
- ⚠️ No automated test of the concurrency invariant (FOR UPDATE); relies on structural assertion + architect reasoning

### Security

- RLS unchanged and documented (M7)
- OU ownership validation in \`change_client_placement\` (RPC rejects OU that doesn't belong to caller's org) — good
- **Permission check mismatch** noted above is the main concern
- \`SECURITY DEFINER\` + \`SET search_path TO 'public', 'extensions', 'pg_temp'\` applied consistently

### Performance

- Partial index \`idx_client_placement_history_ou WHERE is_current = true\` supports \"current placements at this OU\" queries
- \`api.get_client\` gains a LEFT JOIN to \`organization_units_projection\` per placement history row — acceptable given typical history depth (<50 rows) and the projection is pk-lookup
- Backfill is single UPDATE with guards; idempotent

### Recommendation

**Request changes — minor**. The major concern is the \`client.transfer\` enforcement gap: the ADR claims it as the tentpole security decision of the feature but the RPC doesn't check it. Either enforce it in this PR or update the ADR + PR body to be explicit that enforcement lands in PR 2a and the permission is seeded-but-unused until then. Everything else is minor/optional.

The feature work itself is high-quality: architect-reviewed, incrementally structured, well-tested, and the backend foundation (schema + handlers + single-path mutation + lock + read-time enrichment) is solid groundwork for PR 2a/2b.

---

PR #29, opened 2026-04-23
chore(dev): activate api-rpc-readback-pattern follow-up + Phase 0 inventory

lars-tice:
## Summary

- Closes Phase 9 of `client-ou-edit` (mandatory follow-up triggered by PR #27 merge — `e80de9bd`).
- Activates the parked `api-rpc-readback-pattern` feature (`git mv dev/parked/ → dev/active/`) with status frontmatter and activation date updates.
- Populates the Phase 0 RPC inventory tracking table — 19 distinct `api.update_*` / `api.change_*` RPCs catalogued and classified.

### Activation (Phase 9a)
- `dev/active/api-rpc-readback-pattern/` — moved from `dev/parked/`; status PARKED → ACTIVE; `Activated: 2026-04-23`; cross-references to PR 1 proof-of-pattern (`api.update_client`) and PR #27 remediation (`api.change_client_placement`).

### Inventory (Phase 9b)
- 19 RPCs identified via grep across `infrastructure/supabase/supabase/migrations/`. Tracking table in `api-rpc-readback-pattern-plan.md` Phase 0 section classifies each as:
  - **2 EXCLUDED**: `update_client` (PR 1) + `change_client_placement` (PR #27)
  - **1 DONE**: `update_organization_unit` (read-back guard added in `20260221173821`)
  - **1 NEEDS-PATTERN**: `update_schedule_template` (confirmed via grep — has pre-emit existence check but no post-emit projection guard)
  - **15 NEEDS-INSPECTION**: require live `pg_get_functiondef` inspection on the implementation branch
- Live DB query was deferred because the MCP supabase token expired during inventory; the grep + spot-check approach was sufficient for an initial classification table that the implementation branch will refine.

### AGENT-GUIDELINES-compliant Phase 3 doc tasks
- Phase 3 of `api-rpc-readback-pattern-tasks.md` was expanded into six sub-phases (3a–3f) covering: every required TL;DR sub-field for the new ADR, Related Documentation section, AGENT-INDEX update as a multi-step action (keyword table + Document Catalog + cross-validation), `infrastructure/supabase/CLAUDE.md` cross-link, bidirectional ADR linking with `adr-client-ou-placement.md`, and manual frontmatter/link validation (the `Validate Documentation` CI workflow was removed in `430e1c7d`).

### client-ou-edit-tasks.md update
- Phase 9 marked ✅ COMPLETE.
- Current Status updated to reflect: PR 1 merged → Phase 9 done → PR 2a is next.

## What is NOT in this PR

[...]

PR #30, opened 2026-04-23
feat(api): generalize Pattern A v2 read-back guard across 19 RPCs + new ADR

lars-tice:
## Summary

Implements the `api-rpc-readback-pattern` feature (Phases 1+1.6+2+3 of PR 1). Generalizes the projection read-back guard from the 7 pre-existing DONE RPCs (`api.update_client` proof-of-pattern + 6 others) to all 19 in-scope `api.update_*` and `api.change_*` RPCs, and upgrades all 19 from Pattern A v1 (IF NOT FOUND only) to **Pattern A v2** (IF NOT FOUND + race-safe post-emit `processing_error` check on captured event_id).

Closes the silent-failure surface where handlers raised mid-update on existing rows and the RPC returned `{success: true, <entity>: <stale row>}` — a class of bug catalogued in [event-observability.md](https://github.com/Analytics4Change/A4C-AppSuite/blob/main/documentation/infrastructure/guides/event-observability.md) (user invited but no role assigned, organization created but bootstrap incomplete, invitation accepted but status not updated).

Architect-reviewed: [`software-architect-dbc`](https://github.com/anthropics/claude-code) twice — first to choose Pattern A (return-error envelope) over Pattern B (RAISE EXCEPTION + ERRCODE), then to surface the v1 → v2 gap and recommend lockstep retrofit across all RPCs in this same branch.

### Backend (3 migrations)

1. **`20260423060052_api_rpc_readback_pattern.sql`** (Phase 1) — generalizes Pattern A v1 to 11 RPCs (10 NEEDS-PATTERN + 1 COMPLEX-CASE):
   - 5 client sub-entity: `update_client_address`, `_email`, `_funding_source`, `_insurance`, `_phone`
   - 1 organization (BOTH overloads): `update_organization_direct_care_settings` (BREAKING response shape)
   - 3 user: `update_user`, `_phone`, `_notification_preferences`
   - 1 schedule: `update_schedule_template`
   - 1 role (COMPLEX-CASE): `update_role` (composes `roles_projection` + `array_agg(permission_id)` from `role_permissions_projection`)

2. **`20260423062426_add_user_profile_updated_handler.sql`** — creates the missing `handle_user_profile_updated` handler. Discovered during Phase 1 implementation: the router CASE referenced this function since 2026-02-17 but it had never been created. With Pattern A read-back in place, the silent failure surfaced as `{success: true, user: <stale row>}`. Handler reference file added at `infrastructure/supabase/handlers/user/`.

3. **`20260423065747_api_rpc_readback_v2_event_id_check.sql`** (Phase 1.6) — upgrades all 19 single-event RPCs from v1 to v2. v2 adds:
   - Capture: `v_event_id := api.emit_domain_event(...)` (the function already RETURNS uuid)
   - Race-safe post-emit check: `SELECT processing_error INTO v_processing_error FROM domain_events WHERE id = v_event_id; IF v_processing_error IS NOT NULL THEN RETURN error envelope; END IF;`
   - Defense in depth: existing IF NOT FOUND fallback retained
   - `update_role` intentionally NOT retrofitted (uses COMPLEX-CASE 5-second-window pattern appropriate for its multi-emit semantics)

### Frontend (Phase 2)

- `SupabaseDirectCareSettingsService.updateSettings()` adapted to the new `{success, settings}` envelope (was raw jsonb). Backward-compat fallback preserved for any environment where the migration hasn't applied yet.
- `ClientRpcResult` type enriched with optional read-back entity fields (`phone`, `email`, `address`, `policy`, `funding_source`) so consumers can opportunistically use the new RPC response shape without re-fetching.
- 29/29 `DirectCareSettingsViewModel` tests pass.
- Other consumers audited as non-breaking — additive `<entity>` fields are ignored if not used.

### Documentation (Phase 3)

- **NEW ADR** `documentation/architecture/decisions/adr-rpc-readback-pattern.md` (~270 lines, 5 decisions) — formalizes Pattern A with the architect's audit-trail-preservation rationale as the load-bearing constraint. "Pattern A v1 → v2 (Resolved)" section documents the gap closure.
- `documentation/infrastructure/patterns/event-handler-pattern.md` — Required Pattern updated to dual-check v2 example; "Affected RPCs" list inverted from "only org-unit RPCs" to a 7-group table covering all 18 (now 19 with v2) refactored RPCs with explicit exceptions; bidirectional cross-link to ADR.
- `documentation/AGENT-INDEX.md` — 3 new keywords (`api-contract`, `processing-error`, `rpc-readback`) + Document Catalog entry for the ADR.
- `infrastructure/supabase/CLAUDE.md` — existing rule upgraded from "MUST check for NOT FOUND" to "MUST use Pattern A v2 (BOTH checks)" with full code template; explicit "NEVER RAISE EXCEPTION" warning retained.
- `documentation/architecture/decisions/adr-client-ou-placement.md` Decision 2 Enforcement: forward-link to new ADR (proof-of-pattern → general pattern).

## Architectural decisions

Five decisions in the new ADR:

1. **All `api.update_*` and `api.change_*` RPCs MUST perform a post-emit projection read-back** (Pattern A v2 standard form).
2. **Audit-trail preservation (load-bearing)**: handler-driven failures MUST use return-error envelope, NEVER `RAISE EXCEPTION`. Rationale: RAISE EXCEPTION at the RPC layer rolls back the `domain_events` audit row that the BEFORE INSERT trigger just persisted with `processing_error`, destroying the diagnostic evidence the admin dashboard at `/admin/events` and `api.retry_failed_event()` depend on.
3. **Caller-driven failures may use either pattern** (preserve existing per-RPC behavior).
4. **Response shape**: `{success: true, <entity>}` on success; `{success: false, error: 'Event processing failed: ...'}` on failure. HTTP 200 always.
5. **Telemetry convention**: parse `error` prefix `"Event processing failed: "`, expose as dedicated `processingError` ViewModel state for admin "View event in audit log" affordance.

## Total impact

- **19 RPCs** now use Pattern A v2 (race-safe `processing_error` check on captured event_id)
- **1 RPC** uses COMPLEX-CASE multi-event variant (`update_role`)
- **20 function definitions** retrofitted in `20260423065747` (one RPC has 2 overloads)
- **1 missing handler** created (`handle_user_profile_updated`)
- **1 BREAKING response shape change** (direct-care settings) handled with backward-compat in the consumer
- **0 frontend behavioral regressions** (all envelope-consumer contracts preserved)

## Test plan

- [x] `supabase db push --linked --dry-run` — clean for all 3 migrations
- [x] `supabase db push --linked` — applied successfully (one history-repair required during Phase 1.6 due to a heredoc bug; corrected and re-applied atomically)
- [x] Re-dump api schema confirms all 19 RPCs have v_event_id capture + WHERE id = v_event_id post-emit check; `update_role` body unchanged (5-second-window check intact)
- [x] `npm run typecheck` clean (frontend)
- [x] `npm run lint` clean (frontend)
- [x] `DirectCareSettingsViewModel` tests: 29/29 pass
- [x] All 12 relative links in new ADR resolve
- [x] Bidirectional cross-links wired (ADR ↔ event-handler-pattern.md ↔ adr-client-ou-placement.md ↔ infrastructure/supabase/CLAUDE.md)
- [x] AGENT-INDEX keyword entries match TL;DR Key topics
- [ ] **Reviewer to verify**: spot-check 1-2 refactored RPCs in mock mode, confirm response shape includes `<entity>` field on success and `error: 'Event processing failed: ...'` on injected handler failure

## Notes for reviewer

- **Branch chain**: this is the implementation branch for the parked feature `dev/parked/api-rpc-readback-pattern/` activated via PR #29 (Phase 9 of `client-ou-edit`).
- **Migration ordering**: 3 migrations in this PR (`20260423060052`, `20260423062426`, `20260423065747`). All idempotent; all applied to linked dev project; verified via `pg_get_functiondef` re-dumps.
- **Pattern A v1 → v2 history**: v1 shipped first to maintain consistency with the 7 pre-existing DONE RPCs. v2 was triggered by architect follow-up review surfacing the field-level write-through gap (UPDATE-only handlers leak stale rows on partial failure). v2 retrofits both groups in one lockstep migration.
- **`update_role` exception**: kept on COMPLEX-CASE multi-event 5-second-window pattern (not retrofitted to v2) because its multi-emit semantics (1 role.updated + N grants + M revokes) need partial-success awareness that single-event v2 doesn't provide.
- **`update_user` story**: had a missing handler that was never created. Phase 1 surfaced it (commit `461b4929` adds the handler). Phase 1.6 adds v2 race-safe check. Now end-to-end correct.
- **One BREAKING change**: `api.update_organization_direct_care_settings` response shape (raw jsonb → `{success, settings}` envelope). Frontend consumer fixed in `fa0fddd9` with backward-compat fallback for the legacy shape.
- **Future work**: Phase 4 (optional ViewModel simplification — remove redundant `getX()` calls now that RPCs return rows directly) is queued as a separate follow-up PR.

## Dev-docs trail

- Plan: [`dev/active/api-rpc-readback-pattern/api-rpc-readback-pattern-plan.md`](dev/active/api-rpc-readback-pattern/api-rpc-readback-pattern-plan.md)
- Tasks: [`dev/active/api-rpc-readback-pattern/api-rpc-readback-pattern-tasks.md`](dev/active/api-rpc-readback-pattern/api-rpc-readback-pattern-tasks.md)
- Architect agent IDs (cited in commit messages + ADR): pattern-A-vs-B decision, v1→v2 follow-up review


lars-tice (commented):
# Code Review — PR #30: Pattern A v2 Read-back Generalization

## Overview
Generalizes the projection read-back guard (Pattern A) from 7 pre-existing RPCs to all 19 in-scope `api.update_*`/`api.change_*` RPCs, then upgrades all 19 from v1 (IF NOT FOUND only) to **v2** (race-safe `processing_error` check on captured `event_id`). Discovers and fixes a missing handler (`handle_user_profile_updated`) latent since 2026-02-17. Ships a comprehensive ADR documenting the audit-trail-preservation rationale.

**Verdict**: Strong PR — well-architected, well-documented, architect-reviewed twice. Approve with minor non-blocking comments below.

---

## Strengths

- **Audit-trail preservation as load-bearing constraint** is the right call. The `RETURN` vs `RAISE EXCEPTION` distinction is genuinely subtle and the ADR's transaction-rollback explanation (Decision 2) makes a previously-tribal piece of knowledge legible. The cited `20260220185837` recovery example grounds the abstract argument concretely.
- **Race safety of v2 is sound.** `WHERE id = v_event_id` is an indexed PK lookup against the exact emitted row — immune to concurrent emits on the same stream. This is a real improvement over v1's `ORDER BY created_at DESC LIMIT 1`, which could surface a sibling event's error.
- **Defense in depth is preserved correctly** — v1 IF NOT FOUND fallback retained alongside v2 check; both pass = success.
- **Discovery of the missing `handle_user_profile_updated` handler** is exactly the kind of latent bug Pattern A was designed to surface. The fact that the v1 read-back pattern surfaced it (and v2 would have caught it cleanly even without the missing-row case) is the validation story for this whole pattern.
- **Documentation is thorough and bidirectional**: ADR ↔ event-handler-pattern.md ↔ adr-client-ou-placement.md ↔ infrastructure/supabase/CLAUDE.md. AGENT-INDEX keyword updates included. CLAUDE.md rule upgraded to demand BOTH checks.
- **Idempotency**: All migrations use `CREATE OR REPLACE`, safe to re-run. PostgreSQL DDL is transactional, so the 1820-line migration is atomic.
- **Frontend backward-compat**: `SupabaseDirectCareSettingsService` handles both old raw-jsonb and new `{success, settings}` envelope shapes — clean migration path.

---

## Issues

### M1 — Inconsistent IF NOT FOUND lookup: 6 RPCs still use the race-prone query despite having `v_event_id` available

In the v2 migration (`20260423065747`), 6 RPCs use the old race-prone `ORDER BY created_at DESC LIMIT 1` lookup in the IF NOT FOUND branch even though `v_event_id` is now captured in scope. The other 13 RPCs were updated to use `WHERE id = v_event_id` consistently in **both** branches — making this an inconsistency rather than a deliberate choice.

**Affected lines** (`infrastructure/supabase/supabase/migrations/20260423065747_api_rpc_readback_v2_event_id_check.sql`):
- L173 — `update_client_address`
- L243 — `update_client_email`
- L320 — `update_client_funding_source`
- L399 — `update_client_insurance`
- L467 — `update_client_phone`
- L1055 — `update_client` (proof-of-pattern)

**Compare** `update_user` L712-717 and `update_user_notification_preferences` L873-878, which correctly use `WHERE id = v_event_id` in both branches.

**Why it matters**: The IF NOT FOUND branch fires rarely (RLS-denied projection write), so the race window is small — but if a concurrent emit on the same `stream_id`/`event_type` lands first, the wrong `processing_error` gets surfaced. The whole point of v2 is to eliminate that ambiguity. Trivial to fix:

```sql
IF NOT FOUND THEN
    SELECT processing_error INTO v_processing_error
    FROM domain_events WHERE id = v_event_id;  -- use v_event_id, not stream+type
    RETURN jsonb_build_object('success', false,
        'error', 'Event processing failed: ' || COALESCE(v_processing_error, 'unknown'));
END IF;
```

Could land as a follow-up since the post-readback v2 check below catches the same case race-safely — but worth fixing for consistency with the rest of the migration and with the documented standard form in the ADR.

### M2 — `update_role` time-window check has a known weakness; consider captured-event-IDs instead

`update_role` is intentionally NOT retrofitted to v2 because of its multi-emit semantics (1 `role.updated` + N grants + M revokes). The PR description and ADR document this as "5-second processing_error window appropriate for COMPLEX-CASE." But the time-window approach has real issues:

1. The 5-second threshold is arbitrary and could miss errors if processing genuinely took >5s
2. It could surface an unrelated error from another concurrent operation on the same role within the window (e.g. an admin running a script while another admin updates the role via UI)
3. The "success" determination depends on wall-clock time rather than event identity — fragile

A v2-equivalent approach exists: capture each `v_event_id` from the emit calls into a `uuid[]` array, then check if ANY of those specific events have `processing_error`. This would be:
- Race-safe (same indexed PK lookups as single-event v2)
- Correctly scoped to only the events this RPC emitted
- Free of arbitrary time thresholds

```sql
DECLARE
    v_event_ids uuid[] := '{}';
    v_eid uuid;
BEGIN
    -- ... in each emit branch:
    v_eid := api.emit_domain_event(...);
    v_event_ids := array_append(v_event_ids, v_eid);
    -- ...
    SELECT processing_error INTO v_processing_error
    FROM domain_events
    WHERE id = ANY(v_event_ids) AND processing_error IS NOT NULL
    LIMIT 1;
END;
```

Not a blocker — leaving the existing pattern is defensible per the ADR — but the PR's reasoning ("partial-success awareness that single-event v2 doesn't provide") doesn't justify the time-window approach over a captured-IDs approach. Worth filing as a follow-up.

### m3 — `api.update_user` has a pre-existing race on `stream_version` calculation

L682-684 of v2 migration:
```sql
SELECT COALESCE(MAX(stream_version), 0) + 1 INTO v_stream_version
FROM public.domain_events
WHERE stream_id = p_user_id AND stream_type = 'user';
INSERT INTO public.domain_events (..., stream_version, ...) VALUES (..., v_stream_version, ...);
```

Pre-existing pattern (correctly preserved by this PR), but two concurrent `update_user` calls against the same user can compute the same `v_stream_version` and collide on the unique `(stream_type, stream_id, stream_version)` constraint. This is a legacy bug from the baseline (commit `d5d7efea` correctly noted it as a caveat), not a regression — but worth filing as a follow-up: migrate `api.update_user` to use `api.emit_domain_event()` like the other RPCs to inherit advisory-lock-protected stream-version assignment.

### m4 — `ClientRpcResult` type is increasingly polymorphic

`frontend/src/types/client.types.ts:696-718` now has 6 optional read-back entity fields (`client?`, `phone?`, `email?`, `address?`, `policy?`, `funding_source?`) plus 7 optional `*_id` fields. At call sites, none of this is type-safe — consumers must know which RPC populates which field. The pragmatic incremental shape is fine for this PR, but a discriminated union per RPC (or generic `ClientRpcResult<T>`) would catch consumer bugs at compile time. Consider for Phase 4.

### m5 — Sensitive data in `processing_error` strings returned to callers

Pattern A v2 surfaces raw `processing_error` text via `error: 'Event processing failed: ' || v_processing_error`. PostgreSQL handler exceptions may include row data, parameter values, or constraint details that could leak PII or schema internals to the caller. Worth confirming: does `process_domain_event()`'s `WHEN OTHERS` block sanitize `SQLERRM` before persisting, or does it pass through raw exception text? If raw, the admin dashboard `/admin/events` is fine (auth-gated), but returning that text to any authenticated caller of an `update_*` RPC widens the exposure. Quick check: does the trigger catch use `SQLERRM` (raw) or filter through a sanitizer?

---

## Nits

- **Verbosity**: Each retrofitted RPC contains ~5 lines of identical boilerplate for the v2 check. ~100 lines of duplication across 20 functions. The ADR's "no RAISE" constraint makes a clean helper hard to extract (a helper can't `RETURN` from the caller), so leaving the duplication is defensible — but a `_check_event_processing_error_or_envelope(p_event_id) RETURNS jsonb` returning either NULL (continue) or an envelope (caller checks and returns) might be worth exploring.
- **Test plan checkbox open**: The "Reviewer to verify spot-check 1-2 refactored RPCs" item is unchecked. Recommend the assignee close it before merge with at least one mock-mode injection test (e.g. drop a NOT NULL on a projection column, call the RPC, confirm `{success: false, error: 'Event processing failed: ...'}`).
- **Migration apply note** ("required `supabase migration repair --status reverted` due to a heredoc bug") suggests the local→remote roundtrip was non-trivial. Worth a one-line postmortem in `dev/active/` so the next person running a 1800-line migration knows what to watch for.

---

## Test coverage

- ✅ TypeScript `npm run typecheck` clean
- ✅ ESLint clean
- ✅ `DirectCareSettingsViewModel` 29/29 pass
- ✅ ADR link validation (12 relative links resolve)
- ⚠️ No new SQL-level tests for the v2 check itself. plpgsql_check via `supabase db lint` covers static validity but not behavior. Consider adding pg_tap tests (or RPC integration tests) that inject a handler failure and confirm the envelope shape — would have caught M1 immediately.

---

## Approval

**Approve with comments**. M1 is a real but minor inconsistency (6 RPCs); fixing it pre-merge is preferred but it can land as a follow-up since v2 below the IF NOT FOUND branch is race-safe regardless. M2/m3 are pre-existing patterns documented as out-of-scope. m4/m5 are forward-looking.

The architectural foundation here is solid and the audit-trail-preservation rationale will pay dividends across future RPCs.

🤖 Review by Claude Opus 4.7 (1M context)

lars-tice:
## PR Review Remediation — commit `8148889e`

All 8 review findings addressed. Architect-reviewed remediation plan (software-architect-dbc agent `ad2e78383cd378c9f`). Full report + per-finding trade-off analysis in `dev/active/api-rpc-readback-pattern/api-rpc-readback-pattern-tasks.md` "PR #30 Review Remediation" section.

### In this commit

**New migration `20260423074238_api_rpc_readback_v2_m1_m2_fix.sql`** (applied to linked project, verified via `supabase db dump --linked --schema=api`):

- **M1 ✅** — 6 RPCs rewritten to use `WHERE id = v_event_id` in IF NOT FOUND fallback (`update_client_address`, `_email`, `_funding_source`, `_insurance`, `_phone`, `update_client`). Post-apply verification: 0 occurrences of the race-prone `ORDER BY created_at DESC LIMIT 1` pattern remain in these functions.
- **M2 ✅** — `update_role` converted to captured-event-id semantics. `v_event_ids uuid[] := '{}'` captures each emit; error lookup uses `WHERE id = ANY(v_event_ids) AND processing_error IS NOT NULL`. Race-safe under concurrent role edits; empty-array no-op case verified to return `{success: true}`.

**Parked as follow-up feature folders** (architect-recommended scope separation):

- **m3 ⏸️ parked** → `dev/active/update-user-stream-version-race/` — `update_user` manual stream_version race; 0 current callers, orthogonal to Pattern A v2 semantics.
- **m4 ⏸️ Phase 4b plan added** → `dev/active/api-rpc-readback-pattern/api-rpc-readback-pattern-tasks.md` — architect-reviewed Option C (separate named types `ClientPhoneResult`, `ClientEmailResult`, etc. + shared `ClientRpcEnvelope` base). Full DoD included.
- **m5 ⏸️ parked** → `dev/active/rpc-error-pii-sanitization/` — Hybrid Option 6: strip `PG_EXCEPTION_DETAIL` at `process_domain_event` trigger + frontend `maskPii()` utility. 6 alternatives enumerated with trade-offs in the context.md; user's display-masking hint captured as Option 4 component.

**Nit resolutions**:

- **N1 ✅** — boilerplate duplication rationale added to ADR Alternatives Considered. plpgsql single-frame `RETURN` limitation prevents a clean helper extraction; duplication accepted at 20 functions, code generator option deferred.
- **N2 ✅** — spot-check procedure below.
- **N3 ✅** — heredoc postmortem added to `api-rpc-readback-pattern-context.md` Implementation Lessons.

### N2 — Spot-check procedure (for reviewer to execute)

Run in Supabase SQL editor or via `supabase db execute --linked` on a non-production project. Verifies Pattern A v2 envelope correctly surfaces handler failures.

```sql
-- 1. Inject forced handler failure via temporary CHECK constraint on a projection
ALTER TABLE client_phones_projection
    ADD CONSTRAINT spot_check_force_fail CHECK (phone_type != '__SPOT_CHECK__');

-- 2. Call the RPC (replace UUIDs with real test values from your org):
SELECT api.update_client_phone(
  p_client_id := '<valid client uuid>',
  p_phone_id := '<valid existing phone uuid>',
  p_phone_type := '__SPOT_CHECK__'
);
-- Expected response:
-- { "success": false,
--   "error": "Event processing failed: new row for relation \"client_phones_projection\"
--             violates check constraint \"spot_check_force_fail\" ..." }

-- 3. Verify domain_events row was INSERTED (trigger preserves audit trail,
--    did NOT roll back because Pattern A v2 uses RETURN, not RAISE EXCEPTION):
SELECT id, event_type, processing_error, processed_at
FROM domain_events
WHERE event_type = 'client.phone.updated' AND processing_error IS NOT NULL
ORDER BY created_at DESC LIMIT 1;
-- Expect: 1 row; processed_at IS NOT NULL; processing_error contains the constraint message.

-- 4. Cleanup:
ALTER TABLE client_phones_projection DROP CONSTRAINT spot_check_force_fail;
```

Repeat for 3 additional RPCs to cover the refactored surface:
- `api.update_client` (proof-of-pattern)
- `api.update_role` (M2 multi-event captured-ids — inject with constraint on `roles_projection` or `role_permissions_projection`)
- `api.update_client_email` (another M1 target)

If all 4 return `{success: false, error: 'Event processing failed: ...'}` and leave a `domain_events` row with `processing_error` populated, the v2 contract + M1/M2 fixes are confirmed end-to-end.

PR #31, opened 2026-04-23
Phase 4a/4b: Narrow Client RPC types + unblock Roles & ClientFields from loadX() refetch

lars-tice:
## Summary

Four commits advancing the Phase 4 cleanup identified in PR #30 review:

1. **Phase 4b (m4 remediation)** — narrow `ClientRpcResult` (flat union of 13 optional fields) into 9 specific named types extending a shared `ClientRpcEnvelope` base. Architect-reviewed (Option C over A/B).
2. **Phase 4a** — audit 29 ViewModels for redundant post-update fetches; fix the 1 unblocked case (`DirectCareSettingsViewModel`); document 5 blocked cases.
3. **Blocker 1 (Roles)** — extend `RoleOperationResult` with `permission_ids?`; propagate role + permissions through `SupabaseRoleService.updateRole`; `RolesViewModel.updateRole` now patches `rawRoles` in place.
4. **Blocker 2 (ClientFields)** — SQL migration extends 2 RPCs to return entity rows; `RpcResult` narrowed via the Phase 4b pattern; `ClientFieldSettingsViewModel.updateCustomField` / `updateCategory` patch lists in place.

All four commits follow the same recipe: narrow the service result type, pass the Pattern A v2 read-back entity through, consume it in the ViewModel, drop the redundant `loadX()` / `getX()` round-trip.

## Commits

| # | Commit | Scope |
|---|--------|-------|
| 1 | `7b05237e` | `feat(types): Phase 4b — narrow ClientRpcResult into 9 specific envelope types` |
| 2 | `44a9f8ed` | `refactor(vm): Phase 4a — remove redundant loadSettings() in DirectCareSettingsViewModel` |
| 3 | `94325acd` | `feat(roles): Blocker 1 — propagate role + permission_ids through updateRole, drop loadRoles refetch` |
| 4 | `31b5c5c5` | `feat(client-fields): Blocker 2 — narrow RpcResult, return entities from update RPCs, drop loadData refetch` |

### Type refactors (consistent pattern)

- **Client domain** (Phase 4b): `ClientRpcEnvelope` base + `ClientUpdateResult`, `ClientPhoneResult`, `ClientEmailResult`, `ClientAddressResult`, `ClientInsuranceResult`, `ClientFundingResult`, `ClientPlacementResult`, `ClientAssignmentResult`, `ClientVoidResult`. Legacy `ClientRpcResult` deleted after grep confirmed zero external refs.
[...]

PR #32, opened 2026-04-23
Blocker 3 (Scope F PR A): user domain typing + Pattern A v2 read-back consumers + Edge Function v10

lars-tice:
## Summary

Completes the user-domain leg of the Phase 4 cleanup wave (Scope F PR A). Architect-reviewed (software-architect-dbc agent `a9dee2ed181895edb`); 4 MUST-FIX + 3 SHOULD-ADD items integrated.

Three work streams in one commit:
1. **Type narrowing** — `UserOperationResult` (flat union of 5 optional fields × 19 methods) → `UserRpcEnvelope` base + 4 specific result types following Option C pattern established in Phase 4b (client domain).
2. **Pattern A v2 read-back consumption** — service + VM wiring for `update_user_phone`, `add_user_phone`, `update_user`, and the `manage-user` Edge Function's `update_notification_preferences` operation.
3. **New documented pattern** — `documentation/frontend/patterns/rpc-readback-vm-patch.md` at the three-domain threshold (Roles / Client Fields / Users).

### Backend (infra-first per commit-ordering invariant)

- **Migration `20260423232531_add_user_phone_pattern_a_v2_readback.sql`** (applied to linked project): extends `api.add_user_phone` to Pattern A v2. Branches `p_org_id IS NULL` to read from `user_phones` (global) vs `user_org_phone_overrides` (org-scoped) — architect's MUST-FIX (handler writes to two different tables). Returns camelCase `phone` via explicit `jsonb_build_object` (not `row_to_json`) so frontend consumers patch in place without a shape-normalizer step.
- **Edge Function `manage-user` v9 → v10** (deployed): `update_notification_preferences` response now includes `deployVersion` + `notificationPreferences`. Version-gated fallback detection.

### Frontend

- **`frontend/src/types/user.types.ts`** — `UserRpcEnvelope` + `InviteUserResult`, `UpdateUserResult`, `UserPhoneResult`, `UpdateNotificationPreferencesResult`, `UserVoidResult`. Legacy `UserOperationResult` marked `@deprecated`. Address methods use `UserVoidResult` with `TODO(PR-B)` — per architect, phantom `UserAddressResult` before backend exists risks premature shape lock-in.
- **19 service method signatures narrowed** across `IUserCommandService`, `SupabaseUserCommandService`, `MockUserCommandService` via Python bulk substitution (Blocker 1+2 recipe). Typecheck gate caught one Mock shape drift on `resendInvitation` (real service returns bare envelope; Mock was incorrectly returning `invitation`).
- **Entity propagation** — 4 service methods now thread the read-back entity through the narrowed result type, with snake_case → camelCase mapping where the RPC uses `row_to_json`.
- **VM in-place patches** — `UsersViewModel.updateUserPhone`, `.addUserPhone`, `.updateNotificationPreferences` use immutable `.map()` / spread; fallback to refetch + `log.warn` when the entity field is missing (version-gated for the Edge Function path).

### Site resolution

| Site | Original audit | Final resolution |
|------|----------------|------------------|
| Site 1 — `updateUserAddress` → `loadUserAddresses` | "Dead code" (service throws) | ⏸️ **Deferred to PR B** — backend RPCs not yet implemented; PR B is a feature-work plan with its own design session |
| Site 2 — `updateUserPhone` → `loadUserPhones` | "LEGITIMATE (dual-scoped)" | ✅ **Fixed** — misclassified. RPC already returns phone entity; Q7 resolved by architect reading `handle_user_phone_updated.sql` (no sibling-clearing logic). In-place patch correct today; primary-phone exclusivity gap logged as `Blocker-3-followup-1`. |
| Site 3 — `updateNotificationPreferences` → `loadUserOrgAccess` | "SAFE TO SIMPLIFY (Edge Function)" | ✅ **Fixed** — Edge Function v10 + VM version-gated consumer |

### Tests (no prior user-service test coverage existed)

- **`MockUserCommandService.envelope.test.ts`** — 7/7 passing. Exercises narrowed envelope contracts on the mock (entity populated on success; undefined on error envelopes).
- **`UserRpcContract.test.ts`** — 11/11 passing. Anti-drift structural tests (architect Q5 MUST-FIX): parses migration SQL from disk; asserts each Pattern A v2 user RPC's `jsonb_build_object(...)` contains expected keys. Structurally prevents the PR #31 test-drift class from recurring for user RPCs.

### Documentation (AGENT-GUIDELINES.md compliance)

- **NEW**: `documentation/frontend/patterns/rpc-readback-vm-patch.md` — the three-domain VM in-place patch pattern (Roles / ClientFields / Users). Full frontmatter + TL;DR + Related Docs + AGENT-INDEX Keyword/Catalog entries.
- **Updated**: ADR rollout history + user-domain envelope-types mapping table; RPC count 19 → 20.
- **Updated**: `AGENT-INDEX.md` — 2 new keyword rows (`vm-patch`, `in-place-update`) + Document Catalog entry; `last_updated` bumped.
- **Updated**: `dev/active/api-rpc-readback-pattern/api-rpc-readback-pattern-tasks.md` — Blocker 3 complete section + 8 follow-up tasks.

### Follow-up tasks (logged in tasks.md)

| ID | Title |
|----|-------|
| Blocker-3-followup-1 | Primary-phone exclusivity invariant (partial unique index + handler logic) |
| Blocker-3-followup-2 | `manage-user` Edge Function fallback removal (acceptance gate: N days / zero `log.warn` fallback events) |
| Blocker-3-followup-3 | Broader RPC-params contract tests (enumerate all `api.*` functions) |
| Blocker-3-followup-4 | `frontend/src/viewModels/users/CLAUDE.md` VM-level docs |
| Blocker-3-followup-5 | `updateUser` optional in-place patch in consumer VMs |
| Blocker-3-followup-6 | Document Edge-Function-vs-SQL-RPC selection as an ADR |
| Blocker-3-followup-7 | Evaluate breaking up `manage-user` Edge Function into individual SQL RPCs |
| PR-B | Site 1 address backend implementation (separate planning session) |

### Also in this commit

An adjacent anti-staleness `.claude/skills/documentation-writing/SKILL.md` bump landed in the same commit (co-authored). It restructures the skill into Core Structure / Anti-Staleness Guard Rails / Pre-Edit Routine sections based on observed doc-drift failure modes in recent git history. Kept bundled since it's meta-docs guard rails directly relevant to this PR's own doc discipline.

## Test plan

- [x] `supabase db push --linked --dry-run` clean → applied
- [x] `supabase db dump --linked --schema=api` — all 4 expected patterns in `api.add_user_phone` body (2 table branches, `IF p_org_id IS NULL`, `'phone', v_phone`)
- [x] `supabase functions deploy manage-user` — v10 deployed
- [x] `npm run typecheck` ✓
- [x] `npm run lint` ✓
- [x] `npm run build` ✓
- [x] `npm run docs:check` — only pre-existing trailing-slash warning unrelated to this PR
- [x] `npm run test` (user tests) — 18/18 passing
- [ ] **Reviewer to verify**: spot-check mobile/web flow — add a user phone in dev mock mode, confirm VM list updates without a visible refetch; edit notification preferences, confirm UI reflects change in place

## Notes for reviewer

- **Branch name** mirrors `feat/phase4-client-rpc-result-typing` (PR #31).
- **PR-B scope** is intentionally separate — address backend needs design decisions (scope model, primary-address semantics, validation rules) that don't fit a "cleanup" PR.
- **Edge Function deploy-window risk** — VM consumers of `update_notification_preferences` check both `deployVersion` and `notificationPreferences` presence for graceful rollback compatibility. Fallback `log.warn` should reach zero in production within days once v10 stabilizes; removal is `Blocker-3-followup-2`.

### DIFF
--- a/.claude/skills/documentation-writing/SKILL.md
+++ b/.claude/skills/documentation-writing/SKILL.md
@@ -1,31 +1,36 @@
 ---
 name: Documentation Writing Guidelines
 description: Guard rails for documentation quality - frontmatter, TL;DR format, AGENT-INDEX updates, and component prop documentation standards in A4C-AppSuite.
-version: 1.0.0
+version: 1.1.0
+last_updated: 2026-04-23
 category: documentation
-tags: [documentation, agent-index, frontmatter, tldr, component-docs, definition-of-done]
+tags: [documentation, agent-index, frontmatter, tldr, component-docs, staleness, drift]
 ---
 
 # Documentation Guard Rails
 
-Critical rules for creating and updating documentation in the A4C-AppSuite `documentation/` directory. For full rules and templates, see `documentation/AGENT-GUIDELINES.md`.
+Critical rules for creating and updating documentation in the A4C-AppSuite `documentation/` directory. The dominant doc-quality failure mode in this repo is **stale content drifting away from the codebase** — most rules below exist to prevent that.
+
+For full structural rules (placement, templates, quality checklist), see `documentation/AGENT-GUIDELINES.md`. This skill layers anti-staleness discipline on top.
 
 ---
 
-## 1. Required YAML Frontmatter
+## A. Core Structure
+
+### 1. Required YAML Frontmatter
 
 Every documentation file MUST start with:
 
 ```yaml
 ---
 status: current        # or aspirational, archived
-last_updated: 2026-02-05  # today's date
+last_updated: 2026-04-23
 ---
 ```
 
-Update `last_updated` on every edit. Mark aspirational content with inline `> **Note**: This feature is not yet implemented.`
+**Rule**: `last_updated` must equal **today's date, exactly** on every edit — no "close enough", no skipping for "minor" fixes. If you touched the file, bump the date. This is the single cheapest signal a future reader has that the doc is fresh.
 
-## 2. Required TL;DR Section
+### 2. Required TL;DR Section
 
 Every file MUST include a TL;DR block immediately after frontmatter:
 
@@ -45,16 +50,18 @@ Every file MUST include a TL;DR block immediately after frontmatter:
 <!-- TL;DR-END -->
 ```
 
-**Common mistake**: Summary longer than 2 sentences. Keep it concise.
+**Staleness rule**: When the body changes scope, gains/loses a section, or shifts focus, the TL;DR **must be resynced in the same edit**. Symptom check: if you added or removed a section, the Summary and Key topics probably need updating. TL;DR-vs-body drift is recurrent — don't let your edit be the next example.
 
-## 3. New Files Must Update AGENT-INDEX.md
+### 3. Keep AGENT-INDEX.md in Sync
 
-When creating a new documentation file, you MUST:
-1. Add keywords to the keyword table in `documentation/AGENT-INDEX.md`
-2. Add the file to the Document Catalog section with summary, keywords, and token estimate
-3. Verify keywords match the TL;DR `Key topics` field
+`documentation/AGENT-INDEX.md` is the agent's primary entry point. It drifts silently when you skip updates. You MUST update it for:
 
-## 4. Component Props: Inline JSDoc Only
+- **New file**: add to keyword table + Document Catalog (summary + token estimate). Verify keywords match TL;DR `Key topics`.
+- **Rename or move**: grep repo-wide for the old path (`grep -r "old-path.md" documentation/ .claude/ *CLAUDE.md`) and update all hits.
+- **Archive**: remove or repoint all entries pointing to the archived doc.
+- **Significant body change**: if scope or keywords shifted, refresh the catalog summary and re-check the token estimate.
+
+### 4. Component Props: Inline JSDoc Only
 
 Document props directly in the TypeScript interface — no external prop documentation files.
 
@@ -71,7 +78,7 @@ interface ButtonProps {
 
 Use `documentation/templates/component-template.md` for component doc structure.
 
-## 5. Definition of Done: `npm run docs:check`
+### 5. Definition of Done: `npm run docs:check`
 
 Before any frontend PR, documentation validation MUST pass:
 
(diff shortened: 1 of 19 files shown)


lars-tice (commented):
## Review — PR #32 (Blocker 3 / Scope F PR A)

Thorough PR. The three-stream decomposition (type narrowing → read-back consumption → pattern documentation) is well-organized, the architect-review surface is visible in both the migration and tests, and the anti-drift contract tests (`UserRpcContract.test.ts`) are exactly the kind of structural assertion that should prevent the PR #31 class of drift. Most of my feedback below is around one meaningful correctness gap in the Edge Function path and a couple of documentation/testing polish items.

### Overview

- Narrows `UserOperationResult` (flat union, 5 optional × 19 methods) → `UserRpcEnvelope` base + 5 per-method result types. Clean Option C pattern consistent with Blocker 2.
- Threads Pattern A v2 read-back through `updateUserPhone`, `addUserPhone`, `updateUser`, and the `manage-user` Edge Function's `update_notification_preferences` operation.
- Adds a new documented pattern (`rpc-readback-vm-patch.md`) at the three-domain threshold.
- Migration `20260423232531` correctly branches on `p_org_id IS NULL` between `user_phones` and `user_org_phone_overrides`, uses explicit camelCase `jsonb_build_object` (no adapter), and preserves the param list byte-for-byte.

---

### SHOULD-ADDRESS

**1. `update_notification_preferences` Edge Function has a silent-failure gap that SQL Pattern A v2 RPCs don't have.**

`infrastructure/supabase/supabase/functions/manage-user/index.ts:445-460` — the v10 response unconditionally echoes the submitted prefs with `success: true` immediately after `emit_domain_event` returns. There is no guard that checks `processing_error` on the captured `eventId`.

The SQL Pattern A v2 RPCs (e.g. `api.add_user_phone` in the new migration, line 109-118 and 127-134) *do* SELECT `processing_error FROM domain_events WHERE id = v_event_id` and convert that into a `{success: false, error: 'Event processing failed: …'}` envelope. The Edge Function path skips this — meaning if the handler (`handle_user_notification_preferences_updated`) catches an error into the event row's `processing_error` column, the frontend receives `{success: true}` with echoed prefs and patches the VM in place with state that was never persisted.

This is the exact class of silent stale-state bug that the ADR and Pattern A v2 are designed to prevent, and it's worse here than a pre-Pattern-A refetch would be — the refetch would at least reveal the divergence on next read.

Two ways out:
- (Preferred) Add a `processing_error` check after emit in the Edge Function. Small: one `supabaseAdmin.from('domain_events').select('processing_error').eq('id', eventId).single()` round-trip.
- Or explicitly document — in both the Edge Function comment and `rpc-readback-vm-patch.md` — that Edge Function Pattern A v2 consumers accept this silent-failure window as a tradeoff and list the mitigation (cadence of post-write verification, alerting on handler `processing_error` rows, etc.).

Given that `Blocker-3-followup-7` is already "Evaluate breaking up `manage-user` Edge Function into individual SQL RPCs", this gap is arguably a reason to bump that task up the queue or at minimum flag it in the follow-up's motivation.

**2. Factually incorrect comment in the Edge Function.**

`infrastructure/supabase/supabase/functions/manage-user/index.ts:449-454`:

> The handler updated `user_org_access.notification_preferences` from this same JSONB…

The handler actually writes to `user_notification_preferences_projection` (see `infrastructure/supabase/handlers/user/handle_user_notification_preferences_updated.sql`). `user_org_access` is the frontend VM shape, not a DB table. Small fix; matters because this comment is load-bearing for the "echo == read-back" argument and the wrong table name makes the claim unverifiable by a reader.

While fixing, note also that the handler applies COALESCE defaults (`email → true`, `sms.enabled → false`, `inApp → false`), so "echo is equivalent to read-back" is only strictly true when the client submits all fields. In practice the frontend always does (per the type), but the comment should either acknowledge that precondition or switch to an actual read-back.

**3. Absent tests for the `SupabaseUserCommandService` snake_case → camelCase mapping.**

`MockUserCommandService.envelope.test.ts` covers the Mock path well, and `UserRpcContract.test.ts` covers the SQL side. But the non-trivial mapping layer in `SupabaseUserCommandService.updateUserPhone` (lines 971–1003, ~15 field mappings with Date parsing and null handling) and `updateUser` (lines 580–600, similar) has no direct test coverage. This is where PR #31 drift would most likely re-emerge, since a handler could ship a new field in `row_to_json(v_row)::jsonb` without the TS mapper picking it up.

A small unit test that stubs `supabaseService.apiRpc` with a canned snake_case response and asserts the camelCase mapping would close that gap. Not a blocker for this PR but should probably be `Blocker-3-followup-8`.

---

### NITS

**4. `UserOperationResult` deprecation lifecycle.**

`frontend/src/types/user.types.ts:898-909`:

> Scheduled for deletion in the same PR after consumer audit.

A repo-wide grep for `UserOperationResult` outside `user.types.ts` shows no remaining consumers. Either delete now (consistent with the "same PR" intent) or update the comment to name the actual deletion gate (e.g. "delete in Blocker-3-followup-N" or "retained for external package compatibility until vX"). Leaving a deprecated type with "scheduled for deletion in this PR" wording in a merged PR is the kind of stale intent-marker the new SKILL.md warns against.

**5. Hardcoded counts in `rpc-readback-vm-patch.md`.**

Ironic given the SKILL.md update in the same commit explicitly warns against hardcoded inventory counts (rule 8). The new doc lists "Three domains currently use this pattern" and the "Existing implementations" table enumerates specific VM method names. Three is small enough that explicit enumeration is fine, but the table will be the first thing to drift when a fourth domain lands. Either date-stamp the table (`(as of 2026-04-23)`) or add a one-liner linking to a grep recipe that finds all callers. Minor.

**6. `UserFormViewModel.submit` returns a three-arm union.**

`frontend/src/viewModels/users/UserFormViewModel.ts:897`:

```ts
Promise<InviteUserResult | UpdateUserResult | UserVoidResult>
```

Not wrong, but the `UserVoidResult` arm is only reachable in theory (both mode branches invoke methods that return the first two). If a consumer tries to narrow on `result.user` it will pass TS but the third arm is phantom. Could be dropped from the union, or the mode branches could be rewritten as a discriminated union keyed on `this.mode`. Polish-only.

---

### PRAISE

- **Anti-drift contract tests parsing migration SQL on disk** (`UserRpcContract.test.ts`) are exactly the right mechanism. They'll catch both the "RPC shape changed, TS mapper didn't" and the "mapper changed, RPC didn't" drift direction. The assertion on `IF p_org_id IS NULL` + both `FROM user_phones` and `FROM user_org_phone_overrides` is a particularly sharp catch of the architect's MUST-FIX.
- **Edge Function `deployVersion` field** is a cleaner version-gated fallback than field-presence detection. Reusable idea for other Edge Functions.
- **`log.warn` fallback telemetry** at all three VM consumer sites gives the right signal for structural regression without breaking user flow during rollout.
- **SKILL.md anti-staleness update** bundled in the same commit is the right meta-discipline for a PR that itself adds a new documented pattern. Bundling acknowledged in the description.
- **Migration parameter list preserved byte-for-byte** — no drop-and-recreate, no GRANT churn. Correct CREATE OR REPLACE discipline.

---

### Verdict

Approve with the Edge Function `processing_error` guard (item 1) requested before merge, or with an explicit `SHOULD-ADDRESS` follow-up task created and linked in the PR description. Items 2 and 3 are straightforward. Items 4–6 are polish.

_Reviewer: Claude (Opus 4.7). Happy to write the `processing_error` guard patch if you want._

lars-tice:
## Remediation — 2 commits pushed

All 6 review items + architect MUST-FIX / SHOULD-ADD items addressed. Architect-reviewed: `software-architect-dbc` agent `a060ef3faaa5b630c`.

### Commit A — Edge Function v11 (infra-first)

**`e9a39a21`**: `feat(edge-function): manage-user v11 — real Pattern A v2 read-back + org_id audit metadata`

| Review item | Resolution |
|-------------|------------|
| **Item 1 (SHOULD-ADDRESS)** | ✅ v10 echo replaced with genuine Pattern A v2: (1) `SELECT processing_error FROM domain_events WHERE id = eventId` (race-safe PK lookup), (2) `SELECT` 4-column projection row from `user_notification_preferences_projection`. Either check failing produces `{success: false, error: 'Event processing failed: ...'}`. |
| **Item 2 (SHOULD-ADDRESS)** | ✅ Factually-incorrect comment about `user_org_access` removed — replaced with correct reference to `user_notification_preferences_projection` + paired drift-guard comment added to the handler SQL reference file so mismatches become code-review-visible. |
| Architect Q1 SHOULD-ADD | ✅ Edge Function comment cites `adr-rpc-readback-pattern.md:93` (orchestration-tier clause) |
| Architect Q9.2 SHOULD-ADD | ✅ NOT-FOUND on read-back tagged `handlerInvariantViolated: true` (handler UPSERTs) |
| Architect Q9.3 SHOULD-ADD | ✅ **Latent bug fixed** — `buildEventMetadata` call now includes `organization_id` (was missing; violated `infrastructure/CLAUDE.md` Event Metadata Requirements — tenant audit queries were incomplete) |

Deployed via `supabase functions deploy manage-user` — `DEPLOY_VERSION: 'v11-pattern-a-v2-readback'`.

### Commit B — Frontend remediation

**`ffb00780`**: `feat(users): Blocker 3 — PR #32 review remediation (items 3-6 + architect MUST-FIX Q4/Q7 + SHOULD-ADD Q5/Q8/Q9.1)`

| Review item | Resolution |
|-------------|------------|
| **Item 3 (SHOULD-ADDRESS)** | ✅ NEW test file `SupabaseUserCommandService.mapping.test.ts` — 7/7 passing. Uses `vi.hoisted()` + `vi.mock()` to stub apiRpc with canned responses; tests snake_case→camelCase + Date parsing + error envelope |
| **Item 4 (NIT)** | ✅ Legacy `UserOperationResult` deleted after grep-verified zero external refs; 2 stale doc references in `manage-user.md` reference updated |
| **Item 5 (NIT)** | ✅ Pattern doc hardcoded "Three domains" count replaced with grep recipe + date-stamp (SKILL.md rule 8); 4-row table splitting Users into SQL-RPC + Edge-Function paths |
| **Item 6 (NIT)** | ✅ Reviewer's "phantom arm" claim rebutted — `UserVoidResult` IS reachable via `result = roleResult` on modifyRoles failure (line 969). Inline comment with line-citation added (architect Q8) |
| Architect Q4 MUST-FIX | ✅ VM belt-and-suspenders contract-violation log (`contractViolation: true`) for narrow `success: true && !notificationPreferences` case; existing `!data?.success` short-circuit preserved |
| Architect Q5 SHOULD-ADD | ✅ 2 extra mapper tests: `undefined` optional Date; malformed date-string regression guard |
| Architect Q7 MUST-FIX | ✅ ADR `last_updated` bumped + rollout entry for v11 as first Edge Function Pattern A v2 adopter; AGENT-INDEX audited (no stale `user_org_access` refs found); tasks.md Blocker-3-followup-7 motivation strengthened |
| Architect Q8 SHOULD-ADD | ✅ Line-969 citation in UserFormViewModel.submit union comment |
| Architect Q9.1 SHOULD-ADD | ✅ v11 error-envelope contract test in mapper suite |

### Validation

- `npm run typecheck` ✓
- `npm run lint` ✓
- `npm run build` ✓ (5.72s)
- `npm run docs:check` — only pre-existing trailing-slash warning (unrelated)
- **User-domain tests: 25/25 passing across 3 files**
  - `MockUserCommandService.envelope.test.ts` — 7/7
  - `UserRpcContract.test.ts` — 11/11
  - `SupabaseUserCommandService.mapping.test.ts` (NEW) — 7/7
- Grep audits clean (UserOperationResult, user_org_access-as-prefs-table)

### Thank-you notes

- **Item 2 factual correction**: you caught a genuine bug in the comment block. Fixing it properly (via real read-back) eliminated the echo-vs-read-back caveat entirely — cleaner than either documenting the caveat or just fixing the comment.
- **Item 1 framing**: the "silent-failure gap" characterization was precise. v11's two-step check (processing_error + read-back) matches the SQL Pattern A v2 corpus exactly, so the Edge Function now joins the corpus instead of being an exception to it.

### Follow-up (documented, not blocking)

**Blocker-3-followup-7** (evaluate breaking up `manage-user` into individual SQL RPCs) gained two strong data points from this review:
- Reviewer item 1 — Edge Function Pattern A v2 consumers risk regressing this silent-failure class
- Architect finding — a SQL RPC wrapper would be **"strictly superior architecturally"** (single-transaction PL/pgSQL read-back; no two-client-call round-trip)

Depends on Blocker-3-followup-6 (Edge-Function-vs-SQL-RPC ADR) landing first.

PR #37, opened 2026-04-24
fix(users): show toast on notification-preferences save

lars-tice:
## Summary

UI verification after PR #36 merged showed the notification-prefs save silently persists with no user-visible confirmation. Root cause is pre-existing: `handleSaveNotificationPreferences` in `frontend/src/pages/users/UsersManagePage.tsx` set local state on success/error but never called Sonner's `toast`. The rest of the app uses the Sonner `<Toaster />` mounted in `App.tsx:65` (e.g. `ResetPasswordPage.tsx:144`).

Not a regression from the RPC extraction — the save has always worked, just without feedback — but #36's MCP-based end-to-end verification confirmed round-trip works, which is how the UX gap surfaced.

## Scope (post architect-review revisions)

After architect review (`software-architect-dbc`, [comment](https://github.com/Analytics4Change/A4C-AppSuite/pull/37#issuecomment-4316544213)), reverted commit `2ca9d951` (hook extraction + test scaffolding + Playwright UAT — 352 lines) and replaced with the architect's preferred minimal inline change. The hook had one caller, used `useCallback` with a fresh-object-literal dep that defeated memoization, and its "test coverage" actually covered a pure function rather than the hook itself. Architect's call: *"this is a UX fix that needs 8 lines of code, not 352."*

Final branch is **2 commits**:

| Commit | Lines | Purpose |
|---|---|---|
| `5ed147c2` | -363/+38 | Revert hook extraction + tests + Playwright config |
| `49c984db` | +157/-15 | VM routing + error sanitization + drop double-announce + install `@testing-library/react` |

## Files changed

- `frontend/src/pages/users/UsersManagePage.tsx` — `handleSaveNotificationPreferences` rewritten:
  - Routes through `viewModel.updateNotificationPreferences()` instead of `getUserCommandService()` directly. VM owns the Pattern A v2 in-place patch on `userOrgAccess.notificationPreferences` (`UsersViewModel.ts:1794–1818`) plus `isSubmitting` state. Bypassing the VM stranded that method as dead code and created a parallel write path — same dual-write CQRS-compliance regression class the codebase already cleaned up once.
  - Sanitizes RPC error envelope: `Event processing failed:`-prefixed strings carry handler internals (constraint names, processing_error). Show a generic friendly fallback to the user; the raw error stays in the VM's `log.warn` for debugging.
  - Drops `setOperationError` on this path. VM's `error` is cleared via `viewModel.clearError()` after we toast so the page's `role="alert"` banner doesn't double-announce alongside the Sonner `role="alert"` toast.
- `frontend/package.json` + `frontend/package-lock.json` — install `@testing-library/react` ^16.3.2. Architect-flagged: two existing hook test files (`useFocusAdvancement.test.tsx`, `useKeyboardNavigation.test.tsx`) had latent broken imports of `renderHook` from a missing dep. The 56 pre-existing test failures on main are unchanged by this install (same count before and after).

## Architect's must-fixes — addressed

1. ✅ PR description updated to disclose full scope (this section).
2. ✅ ViewModel boundary bypass resolved (now routes through `viewModel.updateNotificationPreferences()`).

## Architect's nice-to-haves — addressed

1. ✅ Error-envelope string leak — sanitized.
2. ✅ Double aria-live announcement — banner cleared after toast on this path.
3. N/A — Playwright config duplication (reverted).
4. N/A — E2E selector fragility (reverted).
5. ✅ `@testing-library/react` installed.
6. N/A — hook unmount cleanup (reverted).

## Architect's standards (post-merge, NOT this PR)

1. Toast + persistent-banner double-surface convention → `frontend/CLAUDE.md` or `documentation/frontend/patterns/command-feedback.md`.
2. "Extract-to-hook-to-test" anti-pattern → `frontend/CLAUDE.md` Testing Patterns section.

These are deferred to follow-up cards per the architect's recommendation.

## Test plan

- [x] `npm run typecheck` clean
- [x] `npm run lint` clean
- [x] `npm run build` clean
- [x] Pre-existing test failure count on main = on this branch (56 failed / 441 passed × both)
- [ ] Manual UI: toggle a preference, save → green Sonner toast `"Notification preferences updated"` appears top-right; preference persists after reload.
- [ ] Manual UI: force an error (e.g. revoke user mid-edit) → red Sonner toast with sanitized message; banner does NOT also display the same error.


lars-tice:
## Architect Review — Changes Recommended

_Posted as comment rather than formal review since GitHub blocks self-requested-changes on own PR._

Overall the underlying toast fix is correct and precedent-consistent (matches `ResetPasswordPage.tsx:144` and `OrganizationFormViewModel.ts:514`), but the PR ships materially more than its description admits and the hidden work has architectural problems.

### Scope gap

The PR description covers only commit `301007cc` (8-line inline `toast.success/error` addition). The actual branch contains a second commit `2ca9d951` that adds 352 undisclosed lines:

| File | Lines | Type |
|---|---|---|
| `frontend/src/hooks/useNotificationPreferencesSave.ts` | +95 | new custom hook |
| `frontend/src/hooks/__tests__/useNotificationPreferencesSave.test.ts` | +124 | test (actually tests a pure function, not the hook) |
| `frontend/playwright.users-uat.config.ts` | +55 | new UAT config |
| `frontend/e2e/users-manage-page.spec.ts` | +78 | new E2E spec |

None of this is mentioned in the PR description. The second commit's own message is honest about what it does, but the PR body is not.

---

## Must-fix (blocks merge)

### 1. Update the PR description to disclose the second commit's scope

The current body implies a simple inline-addition change. Reviewers shouldn't have to discover 352 lines of hook extraction + test scaffolding from the diff.

### 2. Resolve the ViewModel boundary bypass

The new hook calls `getUserCommandService().updateNotificationPreferences(...)` directly (`frontend/src/hooks/useNotificationPreferencesSave.ts:53-57`), bypassing `UsersViewModel.updateNotificationPreferences()` at `frontend/src/viewModels/users/UsersViewModel.ts:1756-1818`. The VM method already provides:

- `isSubmitting` / `error` state management in `runInAction`
- Pattern A v2 read-back with in-place `userOrgAccess.notificationPreferences` patch (lines 1794–1818)
- Contract-violation fallback refetch

The hook skips all of this. Effects:

- **Two parallel write paths** for the same operation
- **Dead code** — `UsersViewModel.updateNotificationPreferences` now has no callers (grep confirms)
- **Lost in-place projection patching** — any future consumer reading `viewModel.userOrgAccess.notificationPreferences` will see stale data

This is the same dual-write CQRS-compliance regression pattern the codebase already cleaned up once (`memory/remediation-history-2026-feb-mar.md`). The inline version in `301007cc` had the same flaw — it too called the service directly — but the hook extraction hardens the bypass by lifting it into a named abstraction that looks canonical.

**Preferred remediation**: revert commit `2ca9d951`, keep `301007cc`'s 8-line inline change. Reasoning:

- Hook has exactly one caller, one call site — violates the root CLAUDE.md rule "three similar lines is better than a premature abstraction"
- The stated justification (testability) is backwards: the test file (`useNotificationPreferencesSave.test.ts`) actually tests `performNotificationPreferencesSave` as a pure function. The hook wrapper (`useState` + `useCallback`) has **zero coverage**. Any function in `UsersManagePage.tsx` could be extracted to a module-local helper and tested the same way without inventing a hook
- **`useCallback([params])` stale-memoization bug** at `useNotificationPreferencesSave.ts:51` — `params` is a fresh object literal on every render of `UsersManagePage.tsx:566-571`, so the callback re-memoizes every render, defeating itself. Doesn't produce incorrect behavior (fresh params are correct), but wastes the `useCallback` entirely
- `@testing-library/react` is NOT installed (`npm ls @testing-library/react` is empty), though other hook test files already import `renderHook` from it (`useFocusAdvancement.test.tsx:2`, `useKeyboardNavigation.test.tsx:2` — latent broken imports). The right fix for hook testability is install the harness, not invent per-call-site hooks

**Alternate remediation if the hook stays**: route through `viewModel.updateNotificationPreferences()` not the service; destructure `userId`/`orgId`/`onError`/`onSuccess` at the top of the hook so `useCallback` deps are primitives; add an `isMountedRef` or `AbortController` guard; add a real hook test using `renderHook` (requires installing `@testing-library/react`).

---

## Nice-to-have (follow-up)

1. **Error-envelope string leak.** `useNotificationPreferencesSave.ts:64-66` passes `result.error` verbatim to `toast.error`. The test at `useNotificationPreferencesSave.test.ts:80-89` codifies surfacing `'Event processing failed: handler column mismatch'` to the end user. That's poor UX and leaks handler internals. Sanitize `Event processing failed:`-prefixed strings to a user-friendly message; keep the raw envelope in `log.error` (already done at line 69). The existing `setOperationError` path had the same flaw — not a regression — but the new test locks the leaky behavior in as expected.
2. **Double aria-live announcement.** The existing `<div role="alert">` error banner at `UsersManagePage.tsx:614` plus Sonner's default error `role="alert"` toast means screen readers may announce the same error twice. Minor but worth a pass.
3. **Playwright config duplication.** `playwright.users-uat.config.ts` is a near-line-for-line duplicate of `playwright.uat.config.ts` — only the port (3459 vs 3458), `testMatch`, and output folder differ. No `npm` script added (`package.json` has no `test:e2e:users-uat`), so CI can't run it without follow-up wiring. Either add the script + CI hook in this PR or explicitly note "manual-only" in the description.
4. **E2E selector fragility.** `frontend/e2e/users-manage-page.spec.ts:51` uses `page.locator('[class*="cursor-pointer"]').filter({ hasText: /@/ })` — a Tailwind-class-substring selector, brittle against class-order changes. Add a `data-testid` to the user row.
5. **Install `@testing-library/react`.** The two existing hook test files that import `renderHook` are latent broken; fixing this unblocks proper hook testing and removes the motivation for extracting logic into pure functions just to test it.
6. **Hook cleanup.** If the hook stays, guard `setIsSaving(false)` against unmount-during-pending (React 18 warns, doesn't crash, but still).

---

## Standards worth codifying (post-merge, not this PR)

1. **Toast + persistent-banner double-surface convention.** This is now de facto at three call sites (`ResetPasswordPage.tsx:144`, `OrganizationFormViewModel.ts:514`, and this PR). Not codified. Proposed: add a short "User feedback on command results" section to `frontend/CLAUDE.md` or a new `documentation/frontend/patterns/command-feedback.md`.
2. **"Extract-to-hook-so-I-can-unit-test-it" as an anti-pattern.** This PR surfaces a recurring tension: MobX ViewModels are the intended test target, but a missing `@testing-library/react` dep pushed the author toward a hook extraction to get around it. The right fix is install the harness. Proposed: `frontend/CLAUDE.md` Testing Patterns section — short form: "Test ViewModels as the unit of MobX logic. Do not extract page-level handlers into bespoke hooks purely to work around a missing test harness — install the test harness."

---

## Verdict

**APPROVE WITH REVISIONS** on the toast fix itself.

**Recommend changes** on the current branch until (1) the PR description is updated and (2) the VM bypass is resolved — preferred path is `git revert 2ca9d951` and ship the 8-line inline change, since this is a UX fix that needs 8 lines of code, not 352.

🤖 Architect review via `software-architect-dbc` agent (Claude)

lars-tice:
## Re-review — Approve

All architect must-fixes addressed cleanly. Verdict flipped from Recommend-Changes → **APPROVE**.

### Must-fixes — verified addressed

**1. PR description updated** ✅ — now accurately reflects the 2-commit (revert + revisions) structure with explicit checklist against the original review comment.

**2. VM boundary bypass resolved** ✅ — `UsersManagePage.tsx:575-579` now calls `viewModel.updateNotificationPreferences({ userId, orgId, notificationPreferences })`. This routes through the VM which:

- Manages `isSubmitting`/`error` in `runInAction` (`UsersViewModel.ts:1764-1767`)
- Performs the Pattern A v2 in-place patch on `userOrgAccess.notificationPreferences` (`UsersViewModel.ts:1794-1818`) with contract-violation refetch fallback
- Uses `result.notificationPreferences ?? preferences` in the page (line 581) to show the server read-back value with an optimistic fallback

### Nice-to-haves — verified addressed

| # | Item | Status |
|---|---|---|
| 1 | Error-envelope sanitization | ✅ `UsersManagePage.tsx:587-591` — strings prefixed `Event processing failed:` replaced with a generic friendly message; raw form retained in VM's `log.warn` (`UsersViewModel.ts:1783`) |
| 2 | Double aria-live fix | ✅ `viewModel.clearError()` on failure path clears `viewModel.error`; banner at `UsersManagePage.tsx:643` is fed from `(viewModel.error \|\| operationError)`, so with VM error null, banner doesn't re-announce |
| 3 | Playwright config duplication | N/A — reverted with `2ca9d951` |
| 4 | E2E selector fragility | N/A — reverted |
| 5 | `@testing-library/react` install | ✅ `package.json` +1, `package-lock.json` +138 |
| 6 | Hook unmount cleanup | N/A — reverted |

### One tiny residual — optional nit

`UsersManagePage.tsx` old code cleared `operationError` (local `useState` at line 116) at the start of every save attempt via `setOperationError(null)`. The diff removes that clear. Effect: if the user has a stale `operationError` from a *different* failed operation (e.g., a failed `deactivate`), then hits Save Preferences successfully, the old error banner remains visible alongside the new success toast.

Minor regression from the page's pre-PR behavior. One-line fix if desired: restore `setOperationError(null);` at the start of the try block (before the `viewModel.updateNotificationPreferences` call). Not a blocker — every other handler on the page (`deactivate`, `reactivate`, `resend invitation`, etc.) already follows the "clear at start of own handler" pattern, so restoring the clear here matches page convention.

### CI note

`gh pr checks 37` shows no checks reported. Expected — the three PR-gated workflows in this repo (`Deploy Database Migrations`, `Deploy Temporal Workers`, `Edge Function ADR Citation`) are all backend-oriented. Frontend-only changes have no PR-gated CI; merge gating relies on the manual `npm run typecheck / lint / build` loop, which the PR body confirms clean. The 56 pre-existing test failures on main are unchanged by this branch.

### Summary

PR went from **363 additions across 5 files** (with scope gap + VM bypass) to **160 additions across 3 files** (all scope disclosed, VM routing correct, error sanitization in place). The 138-line `package-lock.json` growth is the expected side effect of adding `@testing-library/react`.

**Ready to merge.**

🤖 Architect re-review via Claude

PR #39, opened 2026-04-25
refactor(invitation): extract invite-user revoke → api.revoke_invitation RPC

lars-tice:
## Summary

- Second extraction under [adr-edge-function-vs-sql-rpc.md](documentation/architecture/decisions/adr-edge-function-vs-sql-rpc.md) — moves `invite-user revoke` op out of the Edge Function and onto a direct `api.revoke_invitation` SQL RPC call. Inventory row #7 moves `candidate` → `extracted`.
- Migration **DROP + CREATE** widens the return type `boolean → jsonb` and absorbs the Edge Function's gates (`user.create` permission, `access_blocked` JWT-claim guard, JWT-sourced `org_id`). `GRANT EXECUTE` to `authenticated` only.
- **Audit-trail improvements (two)**: prior service-role path emitted `invitation.revoked` events with `event_metadata.user_id = null` AND no correlation_id linkage to the original invitation. Direct authenticated RPC populates the actual revoker's id; lookup-and-reuse pattern propagates `correlation_id` from `invitations_projection` (per `infrastructure/CLAUDE.md` § Correlation ID Pattern + schema commitment at `baseline_v4:12954`).

## Why

`revoke` met none of LB1–LB6 (no auth-user minting, no external API, no workflow forwarding, no bespoke token validation, no cross-tier read orchestration, no pre-user-existence emission). Edge Function added a network hop with no value beyond what an in-DB `SECURITY DEFINER` RPC already provides.

Follows precedent patterns established by [#36](https://github.com/Analytics4Change/A4C-AppSuite/pull/36) (notification-preferences extraction): snake_case wire, `public.has_permission()` for unscoped ops, `public.get_current_user_id()` for caller identity, JWT-sourced `org_id`, DbC `COMMENT ON FUNCTION` block, post-emit `processing_error` check, both TS-types copies regen'd byte-identical.

## Files

- `infrastructure/supabase/supabase/migrations/20260424221149_extract_revoke_invitation_rpc.sql` — new
- `infrastructure/supabase/supabase/functions/invite-user/index.ts` — drop `revoke` case, bump `DEPLOY_VERSION` → `v17-revoke-extracted`
- `frontend/src/services/users/SupabaseUserCommandService.ts` — `revokeInvitation` cuts over to `supabaseService.apiRpc('revoke_invitation', ...)`; 42501 → `'FORBIDDEN'` mapping per in-file precedent (lines 848/955/1052/1115)
- `frontend/src/services/users/__tests__/SupabaseUserCommandService.mapping.test.ts` — 3 new envelope-contract tests for `revokeInvitation`, mirroring the explicit "template for the 4 follow-up extractions" at line 322
- `frontend/src/types/database.types.ts` + `workflows/src/types/database.types.ts` — regen, byte-identical
- `dev/active/invite-user-revoke-to-sql-rpc/{plan,tasks}.md` — Phase 0 findings, locked design decisions D1–D5

**Incidental regen drift**: this PR's `supabase gen types typescript --linked` also picks up a new `graphql_public` schema block in `database.types.ts` (both `frontend/` and `workflows/` copies, kept byte-identical per `infrastructure/supabase/CLAUDE.md`). This was added by an unrelated upstream extension enable — not by migration `20260424221149`. Mentioning here so future archeology with `git blame` on the `graphql_public` lines lands on the right context.

## Verification

**CLI**:
- [x] `npm run lint` (frontend) — clean
- [x] `npm run typecheck` (frontend + workflows) — clean
- [x] `npm run build` (frontend) — clean
- [x] `npx vitest run …mapping.test.ts` — 13/13 pass (10 prior + 3 new revokeInvitation cases)
- [x] TS types regen byte-identical in both consumer copies

**Live DB** (migration applied via `supabase migration repair` + re-push after review-fix amendment):
- [x] RPC shape: `(uuid, text DEFAULT 'manual_revocation') → jsonb`, `SECURITY DEFINER`, EXECUTE on `authenticated` only
- [x] Pre-emit gate: no-JWT call raises `42501 Access denied` (expected)

**UI**:
- [x] Revoke flow toast confirmed end-to-end on dev branch

## Test plan

- [ ] CI: lint + typecheck + plpgsql_check pass
- [ ] CI: Edge Function lint passes (no new files needing ADR citation; `invite-user` is an existing file)
- [ ] Reviewer: confirm migration is idempotent on re-run (DROP IF EXISTS + CREATE)
- [ ] Reviewer: confirm baseline-overload audit (Rule 15) — single signature, no overloads to coexist with

## Review-comment remediation (commit 062286a9)

All 5 review items addressed. Items 1–4 are code/doc changes; Item 5 is this PR description edit:

- **Item 1** (correlation_id reuse) — migration: lookup `invitations_projection.correlation_id`, propagate via `event_metadata.correlation_id`. Now matches `accept-invitation/index.ts:180-188` precedent.
- **Item 2** (TOCTOU acknowledgment) — migration: appended note to DbC `COMMENT` block. Declines reviewer's option 2 (handler tightening) as out of scope; handler is projection-idempotent and duplicate audit rows are intentionally preserved for forensic visibility.
- **Item 3** (hanging "PR" comment) — `invite-user/index.ts:79-81`: dropped the bare PR placeholder, kept the migration-filename + ADR reference.
- **Item 4** (error mapping + tests) — service: 42501 → `'FORBIDDEN'` per in-file precedent; tests: 3 new cases under `revokeInvitation — RPC envelope contract` mirroring the explicit template at `mapping.test.ts:322`.
- **Item 5** (graphql_public regen drift) — this paragraph in the PR description.

## Post-merge bookkeeping

- Archive `dev/active/invite-user-revoke-to-sql-rpc/` → `dev/archived/`
- Append ADR Rollout history entry (inventory row #7 candidate → extracted)
- Update `memory/edge-function-sql-rpc-backlog.md` dependency chain (remove from active list; remaining: `manage-user-delete`, `manage-user-reactivate`, `manage-user-modify-roles`, `manage-user-deactivate-pattern-a-v2-retrofit`)
- Note that PR #39 raises the floor for the remaining 3 candidate extractions: each is now expected to ship correlation reuse + 42501 → FORBIDDEN mapping + envelope-contract unit tests as standard.

### DIFF
--- a/frontend/src/services/users/SupabaseUserCommandService.ts
+++ b/frontend/src/services/users/SupabaseUserCommandService.ts
@@ -255,11 +255,18 @@ export class SupabaseUserCommandService implements IUserCommandService {
 
       if (error) {
         log.error('RPC error in revokeInvitation', { error });
+        if (error.code === '42501') {
+          return {
+            success: false,
+            error: 'Access denied - insufficient permissions',
+            errorDetails: { code: 'FORBIDDEN', message: error.message },
+          };
+        }
         return {
           success: false,
           error: error.message,
           errorDetails: {
-            code: 'UNKNOWN' as UserOperationErrorCode,
+            code: 'UNKNOWN',
             message: error.message,
             context: { postgresCode: error.code, hint: error.hint },
           },
--- a/frontend/src/services/users/SupabaseUserCommandService.ts
+++ b/frontend/src/services/users/SupabaseUserCommandService.ts
@@ -227,9 +227,13 @@ export class SupabaseUserCommandService implements IUserCommandService {
   }
 
   /**
-   * Revoke a pending invitation
+   * Revoke a pending invitation.
    *
-   * Calls the invite-user Edge Function with revoke operation
+   * Calls `api.revoke_invitation` SQL RPC. Extracted from the `invite-user`
+   * Edge Function per `adr-edge-function-vs-sql-rpc.md` — see migration
+   * `20260424221149_extract_revoke_invitation_rpc.sql`. The RPC sources
+   * `org_id` from the JWT, enforces `user.create` permission and the
+   * access-blocked guard server-side, and emits `invitation.revoked`.
    */
   async revokeInvitation(invitationId: string): Promise<UserVoidResult> {
     // Set up tracing context for this operation
@@ -239,29 +243,25 @@ export class SupabaseUserCommandService implements IUserCommandService {
     try {
       log.info('Revoking invitation', { invitationId });
 
-      const client = supabaseService.getClient();
-      const headers = buildHeadersFromContext(tracingContext);
-      const { data, error } = await client.functions.invoke(EDGE_FUNCTIONS.INVITE_USER, {
-        body: {
-          operation: 'revoke',
-          invitationId,
-        },
-        headers,
+      const { data, error } = await supabaseService.apiRpc<{
+        success: boolean;
+        error?: string;
+        eventId?: string;
+        invitationId?: string;
+      }>('revoke_invitation', {
+        p_invitation_id: invitationId,
+        p_reason: 'Revoked by administrator',
       });
 
       if (error) {
-        const errorInfo = await extractEdgeFunctionError(error, 'Revoke invitation');
-        const errorMessage = errorInfo.correlationId
-          ? `${errorInfo.message} (Ref: ${errorInfo.correlationId})`
-          : errorInfo.message;
+        log.error('RPC error in revokeInvitation', { error });
         return {
           success: false,
-          error: errorMessage,
+          error: error.message,
           errorDetails: {
-            code: (errorInfo.code ?? 'UNKNOWN') as UserOperationErrorCode,
-            message: errorInfo.message,
-            context: errorInfo.details ? { details: errorInfo.details } : undefined,
-            correlationId: errorInfo.correlationId,
+            code: 'UNKNOWN' as UserOperationErrorCode,
+            message: error.message,
+            context: { postgresCode: error.code, hint: error.hint },
           },
         };
       }
@@ -273,7 +273,10 @@ export class SupabaseUserCommandService implements IUserCommandService {
         };
       }
 
-      log.info('Invitation revoked successfully', { invitationId });
+      log.info('Invitation revoked successfully', {
+        invitationId,
+        eventId: data.eventId,
+      });
       return { success: true };
     } catch (error) {
       log.error('Error in revokeInvitation', error);
--- a/frontend/src/types/database.types.ts
+++ b/frontend/src/types/database.types.ts
@@ -1306,7 +1306,7 @@ export type Database = {
       retry_failed_event: { Args: { p_event_id: string }; Returns: Json }
       revoke_invitation: {
         Args: { p_invitation_id: string; p_reason?: string }
-        Returns: boolean
+        Returns: Json
       }
       safety_net_deactivate_organization: {
         Args: { p_org_id: string }
@@ -1585,6 +1585,31 @@ export type Database = {
       }
       validate_role_assignment: {
         Args: { p_role_ids: string[] }
+        Returns: Json
+      }
+    }
+    Enums: {
+      [_ in never]: never
+    }
+    CompositeTypes: {
+      [_ in never]: never
+    }
+  }
+  graphql_public: {
+    Tables: {
+      [_ in never]: never
+    }
+    Views: {
+      [_ in never]: never
+    }
+    Functions: {
+      graphql: {
+        Args: {
+          extensions?: Json
+          operationName?: string
+          query?: string
+          variables?: Json
+        }
         Returns: Json
       }
     }
@@ -5859,6 +5884,9 @@ export type CompositeTypes<
 
 export const Constants = {
   api: {
+    Enums: {},
+  },
+  graphql_public: {
     Enums: {},
   },
   public: {
--- a/infrastructure/supabase/supabase/functions/invite-user/index.ts
+++ b/infrastructure/supabase/supabase/functions/invite-user/index.ts
@@ -33,7 +33,7 @@ import {
 import { buildEventMetadata } from '../_shared/emit-event.ts';
 
 // Deployment version tracking
-const DEPLOY_VERSION = 'v16-access-blocked-guard';
+const DEPLOY_VERSION = 'v17-revoke-extracted';
 
 // CORS headers for frontend requests
 const corsHeaders = standardCorsHeaders;
@@ -75,15 +75,19 @@ interface InvitationPhone {
 
 /**
  * Supported operations
+ *
+ * Note: `revoke` was extracted to `api.revoke_invitation` RPC in PR
+ * (migration `20260424221149_extract_revoke_invitation_rpc.sql`) per
+ * adr-edge-function-vs-sql-rpc.md; frontend calls the RPC directly.
  */
-type Operation = 'create' | 'resend' | 'revoke';
+type Operation = 'create' | 'resend';
 
 /**
  * Request format from frontend
  */
 interface InviteUserRequest {
   operation?: Operation;              // Default: 'create'
-  invitationId?: string;              // Required for resend/revoke operations
+  invitationId?: string;              // Required for resend operation
   email?: string;                     // Required for create operation
   firstName?: string;                 // Required for create operation
   lastName?: string;                  // Required for create operation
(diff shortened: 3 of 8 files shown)


lars-tice (commented):
## Review — PR #39: Extract `invite-user revoke` → `api.revoke_invitation` RPC

**Overall**: Solid, well-scoped second extraction under the Edge Function → SQL RPC ADR. Faithful port of the Edge Function gates (auth, `access_blocked`, `user.create`), legitimate audit-trail improvement (revoker `user_id` now populated), idempotent migration shape, and byte-identical type regen across both consumers. Approving in spirit; a handful of non-blocking notes below.

### What this PR does
- Drops the `revoke` case from `invite-user` Edge Function (`v16` → `v17-revoke-extracted`); narrows `Operation` union to `'create' | 'resend'`.
- DROP+CREATE of `api.revoke_invitation`: return type widens `boolean → jsonb`, absorbs the three Edge Function gates server-side, switches caller identity to `public.get_current_user_id()`, sources `org_id` from JWT, adds the post-emit `processing_error` check, GRANTs EXECUTE to `authenticated` only.
- Frontend `SupabaseUserCommandService.revokeInvitation` cuts over to `supabaseService.apiRpc('revoke_invitation', ...)` with the new envelope shape.
- Both `database.types.ts` copies regenerated, byte-identical (return type flips `boolean → Json`, plus an incidental `graphql_public` schema addition — see inline note).

### Strengths
- **Permission semantics match exactly.** Edge `hasPermission(effectivePermissions, 'user.create')` is `arr.some(ep => ep.p === permission)`; SQL `public.has_permission('user.create')` is `EXISTS … WHERE ep->>'p' = p_permission`. Both are presence-only on the `p` field — no behavior change. ✓
- **Pattern A v2 deviation is justified and documented.** The RPC returns outcome-only (no projection entity), so the projection-row read-back step from the canonical pattern is correctly skipped; the captured-`event_id` `processing_error` check is preserved. Plan R1 + the migration's DbC `COMMENT` block both explain this. Consistent with `adr-rpc-readback-pattern.md` for the outcome-only variant.
- **Real audit-trail fix.** Prior service-role path emitted `invitation.revoked` with `event_metadata.user_id = null`. Direct authenticated RPC closes that gap — explicitly called out in the migration header comment. ✓
- **Idempotent migration.** `DROP FUNCTION IF EXISTS` + `CREATE FUNCTION` is safe to re-run. The header explains *why* it's not `CREATE OR REPLACE` (return-type change forces drop). Baseline-overload audit (Rule 15) was performed and documented.
- **DbC `COMMENT ON FUNCTION` block** is excellent — preconditions, postconditions, invariants, error-envelope semantics, and the explicit metadata trade-off note (no `ip_address`/`user_agent`/`request_id`, per CLAUDE.md scoping).

### Notes (non-blocking)
See inline comments at:
- `migrations/…_extract_revoke_invitation_rpc.sql:86` — correlation-id propagation opportunity (improvement, not a regression)
- `migrations/…_extract_revoke_invitation_rpc.sql:72` — pre-existing TOCTOU race, worth one-line acknowledgement
- `functions/invite-user/index.ts:79` — comment placeholder (`in PR\n`) missing the number
- `SupabaseUserCommandService.ts:262` — error-code/correlation-id surface area downgrade + missing test coverage for the new envelope shape
- `frontend/src/types/database.types.ts:1598` — incidental `graphql_public` regen drift, worth a one-liner in the PR description

### Test coverage
The PR's test plan is reviewer-checklist + CI-gates only — no programmatic tests are added. Given the moving parts (envelope-shape contract change + auth-context shift), consider adding at minimum a unit test in `SupabaseUserCommandService.mapping.test.ts` covering: (a) success path with `eventId` propagated to the success log, (b) `data.success === false` envelope path returning `data.error`, (c) `error` (PostgrestError) path with `42501 → 'Access denied'` shape. Existing tests in that file already mock `apiRpc`, so the scaffolding is there.

### Risk summary
- **Breaking-wire-change risk: low** — verified zero non-Edge callers in Phase 0; both Edge case removal and frontend cutover land in the same PR; both type-gen consumers regen'd.
- **Auth-context shift risk: low** — RPC is `SECURITY DEFINER`, all RLS exposure on the projection reads is bypassed inside the function; explicit `org_id IS NULL` and `caller_id IS NULL` guards are present.
- **Backwards-compat: none required** — old `boolean`-returning function had only the deleted Edge Function as a caller.

infrastructure/supabase/supabase/migrations/20260424221149_extract_revoke_invitation_rpc.sql
```diff
@@ -0,0 +1,175 @@
...
+            'invitation_id', p_invitation_id,
+            'reason', p_reason
+        ),
+        p_event_metadata := jsonb_build_object(
+            'user_id', v_caller_id,
+            'organization_id', v_org_id,
+            'source', 'api.revoke_invitation',
+            'reason', p_reason
+        )
+    );
```
lars-tice, line 112: **Correlation-id propagation opportunity (improvement, not a regression).**

This call to `api.emit_domain_event(...)` does not pass `p_correlation_id`. The result is that the `invitation.revoked` event is *not* tied to the original invitation's business-transaction `correlation_id` — same behavior as the prior Edge Function path (which also didn't pass it), so this PR does not regress anything.

However, per `infrastructure/CLAUDE.md` § Correlation ID Pattern ("Updating entity: LOOKUP and REUSE the stored `correlation_id`") and the invitation lifecycle expectation (`user.invited → invitation.resent → invitation.revoked` should share a correlation_id), the canonical fix is one extra read:

```sql
-- (Right after the EXISTS check / before emit)
SELECT correlation_id INTO v_correlation_id
FROM invitations_projection WHERE id = p_invitation_id;

-- … then in the emit call:
p_correlation_id := v_correlation_id,
```

Not a blocker for this PR (the audit-trail-improvement framing in the description is the lead change), but worth either fixing here or queuing as a follow-up tied to the same backlog. The `accept-invitation` Edge Function already does this lookup-and-reuse for the `invitation.accepted` event.

infrastructure/supabase/supabase/migrations/20260424221149_extract_revoke_invitation_rpc.sql
```diff
@@ -0,0 +1,175 @@
...
+    -- (invite-user/index.ts:501-508). Unscoped presence check.
+    IF NOT public.has_permission('user.create') THEN
+        RAISE EXCEPTION 'Permission denied' USING ERRCODE = '42501';
+    END IF;
+
+    -- Existence + status guard (preserved from prior RPC body).
+    SELECT EXISTS(
+        SELECT 1 FROM invitations_projection
+        WHERE id = p_invitation_id AND status = 'pending'
+    ) INTO v_exists;
```
lars-tice, line 80: **TOCTOU note (pre-existing, worth a one-liner acknowledgment).**

The `EXISTS … WHERE status='pending'` check followed by `emit_domain_event` is a classic check-then-act race: two concurrent revokes that both pass the EXISTS check will both emit `invitation.revoked` events. The router's CASE branch for `invitation.revoked` is an unconditional `UPDATE invitations_projection SET status='revoked' WHERE id = …` (no `AND status='pending'` guard in `baseline_v4.sql`), so the second event won't fail — but it will produce a duplicate audit row.

This exact race existed in the prior `api.revoke_invitation` body, so this is not a regression. Either:
1. Acknowledge in the DbC `COMMENT` block ("Concurrent revoke calls may produce duplicate `invitation.revoked` events; the projection update is idempotent."), or
2. Tighten the handler with `WHERE id = v_invitation_id AND status = 'pending'` and rely on `processing_error` to surface the duplicate as a no-op.

Documenting the choice would close the loop.

infrastructure/supabase/supabase/functions/invite-user/index.ts
```diff
@@ -75,15 +75,19 @@ interface InvitationPhone {
 
 /**
  * Supported operations
+ *
+ * Note: `revoke` was extracted to `api.revoke_invitation` RPC in PR
```
lars-tice: **Doc nit: missing PR number in comment.**

The comment reads `Note: \`revoke\` was extracted to \`api.revoke_invitation\` RPC in PR\n  * (migration ...)` — the word `PR` is followed by a line break with no number. Suggest either:
- Inline the number once merged: `… extracted in PR #39 (migration …)`, or
- Drop the bare `PR` to avoid the hanging fragment: `… was extracted to \`api.revoke_invitation\` (migration \`20260424221149_extract_revoke_invitation_rpc.sql\`) per adr-edge-function-vs-sql-rpc.md`.

frontend/src/services/users/SupabaseUserCommandService.ts
```diff
@@ -239,29 +243,25 @@ export class SupabaseUserCommandService implements IUserCommandService {
...
         return {
           success: false,
-          error: errorMessage,
+          error: error.message,
           errorDetails: {
-            code: (errorInfo.code ?? 'UNKNOWN') as UserOperationErrorCode,
-            message: errorInfo.message,
-            context: errorInfo.details ? { details: errorInfo.details } : undefined,
-            correlationId: errorInfo.correlationId,
+            code: 'UNKNOWN' as UserOperationErrorCode,
```
lars-tice: **Error-code surface area downgrade + missing test coverage.**

The prior path used `extractEdgeFunctionError(...)` which produced a structured `errorInfo` (correlation-id-in-message, mapped `code`, `details`). The new path collapses to `code: 'UNKNOWN' as UserOperationErrorCode` and only surfaces `error.code` / `error.hint` in `context`. Two consequences:

1. **`UserOperationErrorCode` cast is sketchy.** Casting the literal `'UNKNOWN'` to the union type is a runtime lie if the union doesn't include `'UNKNOWN'`. If the typecheck passes today only because the union is `string`-flavoured, consider adding `'UNKNOWN'` as an explicit member or mapping the PostgrestError code (`'42501'` → a real `PERMISSION_DENIED` enum) so consumers (toast, telemetry, Sentry tags) get something actionable.
2. **No test guards the new envelope handling.** `SupabaseUserCommandService.mapping.test.ts` already mocks `apiRpc` for similar shapes. Adding three quick cases would lock the contract: success path (with `eventId` logged), `data.success === false` envelope (with `data.error` propagated), PostgrestError 42501 path (with `error.message` propagated to the user-facing string).

Neither is a blocker — both are PR #36 precedent gaps you may already be tracking — but worth a follow-up issue if not addressed here.

frontend/src/types/database.types.ts
```diff
@@ -1595,6 +1595,31 @@ export type Database = {
       [_ in never]: never
     }
   }
+  graphql_public: {
```
lars-tice, line 1598: **Incidental upstream regen drift — worth flagging in PR description.**

This PR's regen also picked up a new `graphql_public` schema block (added by some upstream extension enable, not by this migration). This is fine to land — it's the live shape — but it's not part of this PR's logical change. A one-line note in the PR description ("regen also picks up `graphql_public` schema, unrelated to this migration") will help reviewers of this commit *and* anyone who later does archeology with `git blame` on the `graphql_public` lines.

PR #46, opened 2026-05-05
fix(ui): clear deleted entity-id from URL after successful delete

lars-tice:
## Summary

Reporter (manual UI smoke against \`testorg-20260329\`, 2026-05-05) deep-linked into \`/roles/manage?roleId=X\`, deleted role X, and got an error message after the delete completed. The role was successfully deleted server-side; the URL still contained \`?roleId=X\`; the page's URL→state effect re-fired and tried to re-load the deleted role, surfacing _"Role could not be loaded. Please refresh the page."_

Audit revealed the same bug class in 4 of 5 manage pages. This PR fixes all 4.

## Sequence

1. Page loads with \`?roleId=X\`. URL→state effect calls \`selectAndLoadRole(X)\`. Panel goes to \`edit\` mode.
2. User clicks Delete → confirm. \`handleDeleteConfirm\` calls \`viewModel.deleteRole(X)\`.
3. RPC returns success. Handler resets local state: \`setPanelMode('empty')\`, \`setCurrentRole(null)\`. **URL param \`?roleId=X\` not cleared.**
4. The URL→state effect's deps include \`panelMode\` and \`viewModel.roles.length\`. Both changed. Effect re-fires.
5. Effect reads \`searchParams.get('roleId') === X\` (still the deleted ID), calls \`selectAndLoadRole(X)\`.
6. \`getRoleById(X)\` returns null. Falls into the \`else\` branch: \`setOperationError('Role could not be loaded. Please refresh the page.')\`.

## Fix

In \`handleDeleteConfirm\`, after \`result.success\`, call \`setSearchParams\` with \`{replace: true}\` to remove the now-stale entity-id key from the URL. The URL→state effect will then no-op on its next run because \`searchParams.get('<id-key>')\` is null.

## Pages affected

| Page | Param cleared | Notes |
|---|---|---|
| \`RolesManagePage\` | \`roleId\` | Original report |
| \`UsersManagePage\` | \`userId\`, \`invitationId\` | Also needed to destructure \`setSearchParams\` from \`useSearchParams()\` |
| \`SchedulesManagePage\` | \`templateId\` | |
| \`OrganizationsManagePage\` | \`orgId\` | |

\`OrganizationUnitsManagePage\` was audited and is **not** affected — its URL→state effect gracefully no-ops on \`getUnitById\` returning null instead of surfacing an error.

## Merge gate — User Acceptance Test

A 12-scenario UAT plan is committed to this branch at \`dev/active/manage-pages-url-cleanup-fix/manual-uat-plan.md\`. Highlights:

- Baseline reproduction on \`main\` (optional but recommended)
- Deep-link delete + list-selection delete on each of the 4 affected pages (8 scenarios)
- Failure-path preservation (delete is rejected → URL keeps the entity-id, since the entity still exists)
- Regression check on the unaffected \`OrganizationUnitsManagePage\`
- Bookmark-to-deleted-entity check (explicitly out of scope, verify behavior unchanged)

Pre-merge gate: complete the UAT plan's pass column and attach the report. **Do not merge on automated tests alone** — this repo has no PR-time frontend test workflow today, so local typecheck / lint / unit tests are the only automated signal, and they don't exercise the URL→state interaction this fix targets.

## Local automated checks (before UAT)

| Check | Result |
|---|---|
| \`npm run typecheck\` | ✓ clean |
| \`npm run lint\` (\`--max-warnings 0\`) | ✓ clean |
| \`npm run build\` | ✓ clean |
| Unit tests covering modify_roles surface area (50 cases across 4 files) | ✓ 50/50 pass |

These confirm the fix doesn't break compile-time / type / lint guarantees but **do not** verify the runtime UX. UAT plan above is the gate.

## Out of scope

- Refactoring the URL→state effect to be deletion-aware via the VM rather than waiting for the URL to be reset (defensive engineering, but would touch many call sites).
- Equivalent stale-URL behavior on hard reload of an already-deleted ID (e.g., user bookmarks a deleted role's URL). UAT scenario 12 verifies this stays unchanged; today shows the same error which is arguably correct since the user explicitly navigated to a missing entity.
- Adding a PR-time frontend CI gate (\`typecheck\` + \`lint\` + \`test:run\` + \`build\`). Pre-existing repo gap, larger than this PR; tracked separately.

### DIFF
--- a/dev/active/manage-pages-url-cleanup-fix/manual-uat-plan.md
+++ b/dev/active/manage-pages-url-cleanup-fix/manual-uat-plan.md
@@ -1,0 +1,258 @@
+# Manual UAT — Stale URL param after entity delete (PR #46)
+
+**PR**: https://github.com/Analytics4Change/A4C-AppSuite/pull/46
+**Branch**: `fix/manage-pages-clear-deleted-id-from-url`
+**Surfaced by**: Manual UI smoke test of the modify_user_roles plan, 2026-05-05 — Test 2 (single role removal) reporter landed on `/roles/manage?roleId=...` after deleting a role and saw _"Role could not be loaded. Please refresh the page."_
+
+**Estimated time**: 25–35 minutes if you have admin accounts on the four affected pages prepared (or are willing to seed test entities just for deletion).
+
+---
+
+## What the fix changes
+
+Each of four manage pages, in `handleDeleteConfirm` after `result.success`, now calls `setSearchParams(...)` with `{ replace: true }` to remove the stale entity-id from the URL. This stops the URL→state effect from re-firing on a deleted entity and surfacing the misleading "could not be loaded" error.
+
+| Page | Param now cleared on delete | Test scenarios below |
+|---|---|---|
+| `/roles/manage` | `roleId` | 1, 5, 9 |
+| `/users/manage` | `userId`, `invitationId` | 2, 6, 10 |
+| `/schedules/manage` | `templateId` | 3, 7 |
+| `/organizations/manage` | `orgId` | 4, 8 |
+
+`/organization-units/manage` was audited as **not affected** (its URL→state effect gracefully no-ops on missing entities). It is included in scenario 11 as a regression check — its existing behavior should be unchanged.
+
+---
+
+## Pre-conditions
+
+- Frontend pointed at a dev/staging Supabase project with both PR #44 (modify_roles + M3 registry) and PR #46 (this fix) applied. Local checkout = `fix/manage-pages-clear-deleted-id-from-url`; `npm run dev`.
+- Logged in as a tenant admin in `testorg-20260329` (or any tenant where you can freely create + delete entities).
+- For each page, **at least one disposable entity** to delete:
+  - **Role**: a custom (non-canonical) role you can delete. If none exists, create one first via the Roles list.
+  - **User**: an invited or existing test user. Use a low-privilege test account.
+  - **Schedule template**: a draft or unused template.
+  - **Organization**: this is more invasive — only test on a true dev/staging tenant you're willing to lose. Skip if no disposable org is available; mark scenario 4 as deferred.
+- DevTools open (Network + Console + Elements tabs).
+
+---
+
+## Baseline: reproduce the bug pre-fix (optional but recommended)
+
+Before applying the fix branch, do this once on `main` to confirm you reproduce the same symptom the reporter hit:
+
+1. Check out `main` locally; `npm run dev`.
+2. Navigate to `/roles/manage?roleId=<X>` where X is a deletable role's UUID.
+3. Click Delete; confirm.
+4. **Expected pre-fix**: deletion succeeds in the backend, but the page surfaces _"Role could not be loaded. Please refresh the page."_ The URL still contains `?roleId=<X>`.
+
+If you skip this baseline, the post-fix scenarios are still meaningful — you're just trusting the bug report.
+
+---
+
+## Scenarios
+
+### Scenario 1 — Roles: delete via deep-link
+
+**Steps**:
+1. Navigate to `/roles/manage?roleId=<X>` where X is a deletable role.
+2. Click Delete in the role panel; confirm in the dialog.
+3. Wait for the dialog to close.
+
+**Expected**:
+- The role is removed from the role list.
+- **No error banner.**
+- The URL is now `/roles/manage` (no `?roleId=` query param).
+- Panel returns to empty state.
+
+**DevTools**:
+- Network: a `POST /rest/v1/rpc/delete_role` (or whatever the role-delete endpoint is) returned 2xx.
+- Address bar updates to drop the `roleId` query param.
+- No `setOperationError` console line.
+
+**Pass**: success path completes without the stale error; URL is clean.
+
+---
+
+### Scenario 2 — Users: delete via deep-link
+
+**Steps**:
+1. Navigate to `/users/manage?userId=<U>` where U is a deletable test user.
+2. Use the Delete action (typically in a Danger Zone panel) and confirm.
+
+**Expected**:
+- User removed from list.
+- Toast: "User deleted".
+- URL drops to `/users/manage` (no `?userId=`).
+- No error banner.
+
+**Pass**: as scenario 1, plus `userId` removed from URL.
+
+---
+
+### Scenario 3 — Schedule templates: delete via deep-link
+
+**Steps**:
+1. Navigate to `/schedules/manage?templateId=<T>` where T is a deletable template (no users currently assigned, or the page handles HAS_USERS gracefully).
+2. Click Delete; confirm.
+
+**Expected**:
+- Template removed from list.
+- URL drops to `/schedules/manage`.
+- No error banner.
+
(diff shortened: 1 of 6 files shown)


lars-tice (commented):
## Architectural Review (software-architect-dbc)

**Verdict: PASS WITH MINOR FOLLOW-UPS** — no merge blockers. (Self-PR, so leaving as a review comment rather than an approval.)

The fix is correct, narrow, defensible, and consistent with prior-art patterns already present in these files (the functional-updater `setSearchParams` form with `{ replace: true }` matches the existing status-filter URL update pattern). Bug diagnosis is exact; the DBC postcondition of `handleDeleteConfirm` is now genuinely complete.

### DBC contract framing

**Pre-fix postcondition (incomplete):**
```
panelMode = 'empty' ∧ currentX = null
∧ projection ⊭ {currentX.id}
∧ URL.searchParams[idKey] cleared    ← MISSING
```

**Post-fix postcondition (complete):**
```
panelMode = 'empty' ∧ currentX = null
∧ projection ⊭ {currentX.id}
∧ URL.searchParams[idKey] cleared
∧ ¬(URL→state effect can re-trigger on next render)
```

### Findings by area

1. **Correctness at each call site** — PASS. Placement (after local-state resets, within `result.success` branch, batched same microtask), `{ replace: true }`, functional-updater form, and dependency arrays all correct in all 4 files.
2. **Symmetry across pages** — PASS. Pattern is consistent. `UsersManagePage` clearing both `userId` + `invitationId` is correct defense-in-depth for the unified user/invitation list.
3. **`OrganizationUnitsManagePage` exclusion** — VERIFIED SOUND. That page uses a different (and arguably cleaner) architecture: `?select=` is a one-shot deep-link param, cleared after consumption at `OrganizationUnitsManagePage.tsx:254`; `selectAndLoadUnit` no-ops silently on null at line 273. The bug class is structurally impossible there.
4. **Defensive-engineering tradeoff** — Chosen approach is correct. Hardening the URL→state effect itself (e.g., VM `isDeleted(id)` predicate) would conflate two distinct contracts: (a) deep-link to a missing entity (correct UX: surface error) vs. (b) delete-then-stale-URL re-fire (correct UX: silent cleanup). The PR's "Out of scope" framing on this is the right DBC reasoning.
5. **Cross-tab considerations** — Out of scope, correctly. Pre-existing problem on `main`, not introduced by this PR.
6. **Test plan** — Manual plan adequate as a floor. Recommend a Playwright regression fence (see follow-ups).
7. **No other blockers** — no types / events / RPCs / testids / observability / a11y impact.

### Architecture checklist

| Check | Status |
|---|---|
| CQRS Standards | PASS — purely view-layer |
| Naming Conventions | PASS — uses existing URL keys |
| Design Patterns | PASS — matches existing status-filter pattern in same files |
| data-testid Attributes | N/A — no new UI elements |
| AsyncAPI Event Registration | N/A — no events touched |
| Type Generation (no anonymous types) | PASS — uses platform `URLSearchParams` |
| Observability / Tracing | PASS — existing `log.info` preserved |
| Error Surfacing to UI | PASS — net positive (suppresses misleading stale error; legitimate error path unchanged) |

### Recommended follow-ups (none block merge)

1. **(P2 — recommended)** Add one Playwright spec at `frontend/tests/roles-delete-clears-url.spec.ts` covering the originally-reported flow: `/roles/manage?roleId=X` → click Delete → confirm → assert URL no longer carries `roleId` AND no error banner appears. Optionally parameterize across the four pages. ~30–60 lines; locks in a regression fence for a bug class that already escaped to a tester once.

2. **(P3 — file as separate card, not part of this PR)** `UsersManagePage`'s URL→state effect at `frontend/src/pages/users/UsersManagePage.tsx:152-165` lacks the `panelMode === 'empty'` guard that the other three pages have. The current fix works because all delete-cleared URLs cause the effect's three branches to no-op, but this is a more fragile invariant than the symmetric pages enjoy. Pre-existing fragility — **not introduced by this PR**. Worth a hardening card.

3. **(P4 — cosmetic, optional)** Comment text in `RolesManagePage.tsx:462-465` is more verbose (with reproduction date) than the abridged comments in the other three files. Either keep the canonical case verbose or abridge for strict symmetry; neither is wrong.

### Bottom line

Ready to merge. The minor follow-ups are issue-board work, not merge gates.

lars-tice:
## Architect follow-ups — addressed in commit `bb2145e8`

Posting back per-finding so the resolution trail lives next to the recommendation. All three are addressed; only P2/P4 land in this PR per the architect's own scoping (P3 explicitly deferred to a separate card).

### P2 — Regression fence ✅ (in this PR)

Added at [`frontend/src/pages/__tests__/manage-pages-clear-deleted-id-from-url.test.tsx`](https://github.com/Analytics4Change/A4C-AppSuite/blob/bb2145e8/frontend/src/pages/__tests__/manage-pages-clear-deleted-id-from-url.test.tsx). 8 cases, 2 per affected page, parameterized over the four contracts:

| Page | Keys cleared | Co-resident params preserved |
|---|---|---|
| `RolesManagePage` | `roleId` | `?status=active` |
| `UsersManagePage` | `userId`, `invitationId` | `?mode=edit` |
| `SchedulesManagePage` | `templateId` | `?status=active` |
| `OrganizationsManagePage` | `orgId` | `?status=all` |

Per-page assertions: (a) target key removed from URL, (b) unrelated params preserved, (c) `{replace: true}` honored (location key changes), (d) re-running on already-clean URL is a no-op.

**Substituted Vitest+RTL+MemoryRouter for Playwright.** Rationale documented in the test file's top-level docblock — the existing Playwright fixtures rely on real auth + real entity rows per page per browser-project (disproportionate fixture work for a URL-handling closure), and there's a pre-existing `playwright.config.ts:10` `testDir: './e2e'` vs. actual `frontend/tests/` location mismatch that means existing specs aren't on the gated path today. The contract under test is purely `setSearchParams ↔ React Router URL state`, both faithful in JSDOM. If the Playwright test infra is hardened later (mock auth + fixture seeding), upgrading this to a true e2e regression is a fast-follow.

8/8 pass. Combined with the existing service+component suite, 58/58 in the affected surface area.

### P3 — `UsersManagePage` URL→state effect missing `panelMode === 'empty'` guard ↗ (separate card)

Honored your "**file as separate card, not part of this PR**" framing. Memorialized in `MEMORY.md` Active Backlogs as a deferred work item with this PR's commit as the seed reference. Working title for the card: `users-manage-page-url-effect-empty-guard`. Trigger to seed: bandwidth opens OR a user reports a re-fire bug on `/users/manage`. Pre-existing fragility — **not introduced by this PR**.

### P4 — Comment symmetry ✅ (in this PR)

Abridged the verbose `RolesManagePage.tsx:462-465` comment (4 lines + reproduction date) to the canonical 2-line rule used in the other three files:

```typescript
// Clear stale URL params so the URL→state effect doesn't re-select
// the just-deleted entity and surface a stale "could not load" error.
```

All four call sites now read identically.

---

Verification on `bb2145e8`:
- `npm run typecheck` ✓ clean
- `npm run lint --max-warnings 0` ✓ clean
- Affected unit tests: 58/58 pass

**UAT plan still required as the merge gate** — see [`dev/active/manage-pages-url-cleanup-fix/manual-uat-plan.md`](https://github.com/Analytics4Change/A4C-AppSuite/blob/bb2145e8/dev/active/manage-pages-url-cleanup-fix/manual-uat-plan.md) (12 scenarios, mirrors the modify_user_roles smoke plan structure).

PR #54, opened 2026-05-08
fix(invitation): drop deprecated role column; switch to roles[] contract

lars-tice:
## Summary

- Drops the deprecated `invitations_projection.role` column (tagged DEPRECATED in baseline_v4 line 12918, never populated by any current emitter).
- Migrates the validate-invitation Edge Function response and the frontend `InvitationDetails` type from singular `role: string` to plural `roles: Array<{role_id, role_name}>`, matching the existing JSONB column and the AsyncAPI contracts (which already use `roles[]` exclusively).
- Removes a silent AsyncAPI conformance violation in `accept-invitation` (the `invitation.accepted` event_data was emitting `role: invitation.role` despite the contract for `InvitationAcceptedData` never specifying that field).

## Defect

When a user clicks an invitation email link, the page rendered:

> Invalid Invitation
> Invalid invitation response

Root cause: `api.get_invitation_by_token` returned `role: null` (the singular column was always NULL for current emitters). The frontend's `SupabaseInvitationService` threw on `!data?.role`, even though `role: null` is a legitimate state per the `invite-user` contract ("permissions assigned later" - empty `roles[]`).

## Changes

| Layer | File | Change |
|---|---|---|
| **Migration** | `infrastructure/supabase/supabase/migrations/20260508170054_drop_invitations_deprecated_role_column.sql` | NEW: handler update + RPC DROP/CREATE + defensive backfill + DROP COLUMN |
| **Handler ref** | `infrastructure/supabase/handlers/user/handle_user_invited.sql` | Removed `role` from INSERT |
| **Edge Fn** | `infrastructure/supabase/supabase/functions/validate-invitation/index.ts` | Response shape `role` → `roles[]`; v11-roles-array |
| **Edge Fn** | `infrastructure/supabase/supabase/functions/accept-invitation/index.ts` | Dropped singular fallback + `role` from event_data; v22-roles-array |
| **Frontend** | `frontend/src/types/organization.types.ts` | `InvitationDetails.role` → `roles[]` |
| **Frontend** | `frontend/src/services/invitation/SupabaseInvitationService.ts` | Guard + mapping |
| **Frontend** | `frontend/src/services/invitation/MockInvitationService.ts` | 5 sites |
| **Frontend** | `frontend/src/pages/organizations/AcceptInvitationPage.tsx` | Conditional list render |
| **Frontend** | `frontend/src/viewModels/organization/InvitationAcceptanceViewModel.ts` | Logging field |
| **Workflows** | `workflows/src/scripts/query-dev.ts` | Type + table + CSV |
| **Both** | `*/types/database.types.ts` | Regen (byte-identical) |

## Architect review

Reviewed by `software-architect-dbc` agent during plan mode. Ten findings; all incorporated:
- **#1 BLOCKER**: write-site audit added to migration header (verified `process_invitation_event` `WHEN 'user.invited'` INSERT at baseline_v4:11064 is dead code and already correct).
- **#3 HIGH**: removed legacy `'roleName'` (camelCase) COALESCE arm from new RPC body too.
- **#5 MEDIUM**: `MockInvitationService` site count corrected from 2 to 5.
- **#8 MEDIUM**: defensive `UPDATE` backfill before `DROP COLUMN` (no-op on this DB; safety net for legacy rows).
- Other findings: confirmed in plan or addressed mechanically.

## Deployment status (already applied to dev)

The destructive operations were already executed against the linked dev project as part of the in-band fix flow:

- ✅ Migration applied (`supabase db push --linked`).
- ✅ Both Edge Functions deployed (`supabase functions deploy validate-invitation`, `supabase functions deploy accept-invitation`).
- ✅ `database.types.ts` regenerated (frontend + workflows byte-identical).

This PR is the source-of-truth commit for those already-shipped changes plus the frontend/workflows code that must deploy via CI to take effect on `firstovertheline.com`.

## Verification

**Live wire-level check** (post-deploy):
```bash
$ curl -sS -X POST "$SUPABASE_URL/functions/v1/validate-invitation" \
    -H "Authorization: Bearer $ANON_KEY" -H "apikey: $ANON_KEY" \
    -d '{"token":"<failing-token>"}'
{"valid":true,"orgName":"TestOrg-20260329","roles":[],...}
```

No more `role: null`; `roles: []` is the new shape.

**Local gates** (all passing on this branch):
- `frontend`: typecheck ✓, lint ✓, build ✓, docs:check ✓
- `workflows`: build ✓, 12/12 tests ✓

## Test plan

- [ ] Frontend deploys via `frontend-deploy.yml` on merge.
- [ ] Reload the failing invitation URL on `firstovertheline.com` → page renders org name "TestOrg-20260329", omits the role line (since `roles[]` is empty), shows the auth-method picker. No red "Invalid Invitation" banner.
- [ ] Test a fresh invitation **with** a role assigned → role line renders correctly.
- [ ] DB sanity: `SELECT event_data ? 'role' FROM domain_events WHERE event_type = 'invitation.accepted' ORDER BY created_at DESC LIMIT 5` returns all `false` for new acceptance events.
- [ ] CI gates: rpc-registry-sync workflow stays green (registry already lists `get_invitation_by_token` as `read`; my migration re-issued the same tag).
- [ ] Resume modify_user_roles UAT (Scenarios 1, 3, 4, 6, 7, 8, 9, 10) which was blocked by this defect.

## What this kills

- Dead column in production schema.
- Always-null event_data field (silent AsyncAPI conformance violation).
- COALESCE-from-array fallback in `get_invitation_by_token` whose first argument was always NULL.
- Dead-code fallback in `accept-invitation` defending against an impossible state.
- The data-model lie that "an invitation has one role string."

### DIFF
--- a/.github/workflows/rpc-registry-sync.yml
+++ b/.github/workflows/rpc-registry-sync.yml
@@ -42,12 +42,29 @@ jobs:
         with:
           version: latest
 
+      # `supabase start` auto-applies any migrations in supabase/migrations/
+      # on first init (confirmed by failing log on the prior CI run). Move the
+      # directory aside before start so the container boots with no migrations,
+      # then restore for the phased manual apply below. Without this, the May
+      # 6 postcondition migration runs before our seed step has a chance.
+      - name: Move migrations aside (prevent supabase start auto-apply)
+        working-directory: infrastructure/supabase
+        run: |
+          mv supabase/migrations supabase/migrations.staged
+          mkdir supabase/migrations  # supabase CLI requires the directory to exist
+
       - name: Start local Supabase container
         working-directory: infrastructure/supabase
         run: |
           supabase start
           # Wait for the container to be fully ready (status reports DB URL)
           supabase status
+
+      - name: Restore migrations directory
+        working-directory: infrastructure/supabase
+        run: |
+          rmdir supabase/migrations
+          mv supabase/migrations.staged supabase/migrations
 
       - name: Install psql client
         run: |
--- a/.github/workflows/rpc-registry-sync.yml
+++ b/.github/workflows/rpc-registry-sync.yml
@@ -43,15 +43,25 @@ jobs:
           version: latest
 
       # `supabase start` auto-applies any migrations in supabase/migrations/
-      # on first init (confirmed by failing log on the prior CI run). Move the
-      # directory aside before start so the container boots with no migrations,
-      # then restore for the phased manual apply below. Without this, the May
-      # 6 postcondition migration runs before our seed step has a chance.
-      - name: Move migrations aside (prevent supabase start auto-apply)
+      # on first init AND requires the `api` schema to exist for PostgREST
+      # health checks to pass. Strategy: keep ONLY baseline_v4 in the dir
+      # during supabase start (it creates the `api` schema and the foundational
+      # structures), stage the rest aside. After start succeeds, restore the
+      # rest and apply via the phased manual flow below. Without this, either:
+      #   (a) all migrations apply and the May 6 postcondition fails, or
+      #   (b) no migrations apply and PostgREST health check times out (503).
+      - name: Stage post-baseline migrations aside
         working-directory: infrastructure/supabase
         run: |
-          mv supabase/migrations supabase/migrations.staged
-          mkdir supabase/migrations  # supabase CLI requires the directory to exist
+          mkdir supabase/migrations.staged
+          # Move every migration EXCEPT baseline_v4 to the staged dir
+          for f in supabase/migrations/*.sql; do
+            base=$(basename "$f")
+            if [[ "$base" != "20260212010625_baseline_v4.sql" ]]; then
+              mv "$f" supabase/migrations.staged/
+            fi
+          done
+          ls supabase/migrations/
 
       - name: Start local Supabase container
         working-directory: infrastructure/supabase
@@ -60,11 +70,11 @@ jobs:
           # Wait for the container to be fully ready (status reports DB URL)
           supabase status
 
-      - name: Restore migrations directory
+      - name: Restore staged migrations
         working-directory: infrastructure/supabase
         run: |
-          rmdir supabase/migrations
-          mv supabase/migrations.staged supabase/migrations
+          mv supabase/migrations.staged/*.sql supabase/migrations/
+          rmdir supabase/migrations.staged
 
       - name: Install psql client
         run: |
@@ -98,15 +108,23 @@ jobs:
         run: |
           PSQL_CMD="psql -h 127.0.0.1 -p 54322 -U postgres -d postgres -v ON_ERROR_STOP=1"
 
-          # Boundary timestamp: apply migrations through and INCLUDING this one
-          # before seeding. The postcondition migration 20260506190626 must run
-          # AFTER the seed.
+          # Baseline (20260212010625_baseline_v4.sql) is ALREADY APPLIED by
+          # `supabase start` above (we kept it in the migrations dir so PostgREST
+          # health checks pass). Phase 1 applies post-baseline migrations through
+          # and INCLUDING the boundary timestamp; the postcondition migration
+          # 20260506190626 runs in Phase 3 after the permissions seed.
+          BASELINE_TS="20260212010625"
           BOUNDARY_TS="20260506012315"
 
-          echo "::group::Phase 1 - Migrations through and including ${BOUNDARY_TS}"
+          echo "::group::Phase 1 - Post-baseline migrations through and including ${BOUNDARY_TS}"
           for f in $(ls supabase/migrations/*.sql | sort); do
             base=$(basename "$f")
             ts=${base%%_*}
+            # Skip baseline (already applied by supabase start)
+            if [[ "$ts" == "$BASELINE_TS" ]]; then
+              echo "Skipping $base (already applied by supabase start)"
+              continue
+            fi
             if [[ "$ts" > "$BOUNDARY_TS" ]]; then
               echo "Stopping at $base (will apply after seed)"
               break
--- a/.github/workflows/rpc-registry-sync.yml
+++ b/.github/workflows/rpc-registry-sync.yml
@@ -49,9 +49,83 @@ jobs:
           # Wait for the container to be fully ready (status reports DB URL)
           supabase status
 
-      - name: Apply migrations to local container
+      - name: Install psql client
+        run: |
+          sudo apt-get update -qq
+          sudo apt-get install -y -qq postgresql-client
+
+      # Migrations applied in three phases with a permissions seed step in
+      # between. Rationale:
+      #
+      # 20260506190626_add_role_management_view_ou_implications.sql ends with a
+      # DO $$ ... ASSERT $$ postcondition that requires 6 implication rows to
+      # exist, which depends on permissions_projection being populated. On
+      # dev/prod, permissions were populated via the post-baseline seed file
+      # `infrastructure/supabase/sql/99-seeds/001-permissions-seed.sql` (it
+      # emits permission.defined events that the trigger projects into the
+      # projection). On a fresh CI container, that seed never runs (the CLI's
+      # supabase/seed.sql doesn't exist; the 99-seeds/*.sql files aren't
+      # auto-applied).
+      #
+      # The prior `supabase db reset --local` step blindly ran every migration
+      # in lexical order and failed at 20260506190626 with
+      # "Postcondition violated: expected 6 implication rows, found 0".
+      #
+      # Replacement: drive migrations manually via psql, applying baseline +
+      # early migrations first, seeding permissions, then applying remaining
+      # migrations.
+      - name: Apply migrations and seed permissions in phased order
         working-directory: infrastructure/supabase
-        run: supabase db reset --local
+        env:
+          PGPASSWORD: postgres
+        run: |
+          PSQL_CMD="psql -h 127.0.0.1 -p 54322 -U postgres -d postgres -v ON_ERROR_STOP=1"
+
+          # Boundary timestamp: apply migrations through and INCLUDING this one
+          # before seeding. The postcondition migration 20260506190626 must run
+          # AFTER the seed.
+          BOUNDARY_TS="20260506012315"
+
+          echo "::group::Phase 1 - Migrations through and including ${BOUNDARY_TS}"
+          for f in $(ls supabase/migrations/*.sql | sort); do
+            base=$(basename "$f")
+            ts=${base%%_*}
+            if [[ "$ts" > "$BOUNDARY_TS" ]]; then
+              echo "Stopping at $base (will apply after seed)"
+              break
+            fi
+            echo "Applying $base"
+            ${PSQL_CMD} -f "$f"
+          done
+          echo "::endgroup::"
+
+          echo "::group::Phase 2 - Seed permissions (mirrors dev/prod post-baseline state)"
+          ${PSQL_CMD} -f sql/99-seeds/001-permissions-seed.sql
+          echo "::endgroup::"
+
+          echo "::group::Phase 3 - Apply remaining migrations after ${BOUNDARY_TS}"
+          for f in $(ls supabase/migrations/*.sql | sort); do
+            base=$(basename "$f")
+            ts=${base%%_*}
+            if [[ "$ts" > "$BOUNDARY_TS" ]]; then
+              echo "Applying $base"
+              ${PSQL_CMD} -f "$f"
+            fi
+          done
+          echo "::endgroup::"
+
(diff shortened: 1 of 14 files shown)


infrastructure/supabase/supabase/functions/accept-invitation/index.ts
```diff
@@ -936,7 +939,6 @@ serve(async (req) => {
           org_id: invitation.organization_id,
           user_id: userId,
           email: invitation.email,
-          role: invitation.role,
           accepted_at: new Date().toISOString(),
         },
```
lars-tice, line 987: **BLOCKER — AsyncAPI conformance regression.**

The PR description (and the comment block immediately above this emit) claims `InvitationAcceptedData` does not specify a `role`/`roles` field. That's only half right. From `infrastructure/supabase/contracts/asyncapi/domains/invitation.yaml:362-368`:

```yaml
InvitationAcceptedData:
  required:
    - invitation_id
    - org_id
    - user_id
    - email
    - roles      # <-- REQUIRED
    - accepted_at
```

So:
- Removing `role: invitation.role` here is correct (singular `role` was never in the contract).
- But `roles` is a **required** field in the contract, and this event_data does not include it. After this PR, every `invitation.accepted` event will be missing a required field — a **new** conformance violation that replaces the old one.

The local `roles: RoleRef[]` variable already in scope at line 718 is exactly the right value (per the schema description: "Snapshot of roles from the invitation that were assigned to the user. Empty array = user accepted with no roles"). Fix:

```ts
p_event_data: {
  invitation_id: invitation.id,
  org_id: invitation.organization_id,
  user_id: userId,
  email: invitation.email,
  roles,                 // <-- add this; matches required schema field, allows []
  accepted_at: new Date().toISOString(),
},
```

Also update the comment block above this emit and the PR description — both currently state the contract has no `role`/`roles` field, which will mislead the next reader. Recommend a Deno test (or a domain_events SQL check appended to the test plan) that asserts `event_data ? 'roles'` is true for new acceptance events, symmetric to the existing `? 'role'` check.

infrastructure/supabase/supabase/functions/accept-invitation/index.ts
```diff
@@ -707,16 +707,15 @@ serve(async (req) => {
...
 
-    // Get roles from new format (roles array) or legacy format (single role string)
+    // Roles to assign on acceptance. The legacy singular `role` fallback was
+    // removed 2026-05-08 along with the deprecated invitations_projection.role
+    // column (migration 20260508170054). Empty array is a legitimate state
+    // ("permissions assigned later" — the invite-user contract supports this).
     interface RoleRef {
       role_id: string | null;
       role_name: string;
     }
```
lars-tice, line 723: **Minor — type tightens against actual data shape.**

This inline `RoleRef.role_name: string` declares non-null, but every other layer of this code path types it as `string | null`:
- `frontend/src/types/organization.types.ts` `InvitationDetails.roles[].role_name: string | null`
- `validate-invitation/index.ts` `InvitationRoleRef.role_name: string | null`
- `MockInvitationService.ts` `MockInvitationData.roles[].role_name: string | null`
- The `roles` JSONB in `invitations_projection` is unconstrained at the DB level.

The role-assignment loop at line 728 only guards on `!roleId`; if `role_name` arrives as null (legitimate per the upstream types), it will be passed as `p_role_name: null` to `api.get_role_by_name` and the user will see `role_lookup_failed` even though the more accurate failure is "invitation contained a role with no identifying info." Suggest:

```ts
interface RoleRef {
  role_id: string | null;
  role_name: string | null;  // align with rest of stack
}
```

and expand the guard to `if (!roleId && !roleName) { /* skip + log warn */ }` (or fail with a clearer error).

infrastructure/supabase/supabase/migrations/20260508170054_drop_invitations_deprecated_role_column.sql
```diff
@@ -0,0 +1,170 @@
...
+-- Step 3: Defensive backfill - preserve any legacy role-only data into roles[]
+-- before dropping the column. No-op on a clean DB; protects against any
+-- unknown legacy rows that have role IS NOT NULL with empty roles[].
+-- -----------------------------------------------------------------------------
+UPDATE public.invitations_projection
+SET roles = jsonb_build_array(
+  jsonb_build_object('role_id', NULL, 'role_name', role)
+)
+WHERE role IS NOT NULL
+  AND (roles IS NULL OR roles = '[]'::jsonb);
```
lars-tice, line 165: Backfill is well-scoped (no-op on a clean DB, safety net otherwise) and uses `jsonb_build_object` correctly. Two small notes for posterity, neither a blocker:

1. The architect-reviewed write-site audit at the top is solid, but consider also asserting it at runtime: `DO $$ BEGIN ASSERT (SELECT count(*) FROM invitations_projection WHERE role IS NOT NULL AND (roles IS NULL OR roles = '[]'::jsonb)) = 0; END $$;` after the UPDATE would loudly catch a silent backfill miss before the DROP COLUMN. Postcondition style is the same idiom used in `20260506190626_add_role_management_view_ou_implications.sql`.
2. The `role_id: NULL` in the synthesized JSONB will make a downstream `accept-invitation` flow fall into the `get_role_by_name` lookup branch — fine, but if any of these legacy rows exists for a role that no longer exists by name, acceptance will fail with `role_lookup_failed`. Acceptable since the audit confirms zero such rows exist; just want it documented for the next person.

lars-tice (commented):
Found one BLOCKER plus a few smaller items. Summary:

- **BLOCKER (AsyncAPI conformance regression)** in `accept-invitation/index.ts` — see inline comment on the `invitation.accepted` emit. The PR removes the singular `role` (good — was never in the contract), but does NOT emit the **required** `roles` field. Per `infrastructure/supabase/contracts/asyncapi/domains/invitation.yaml:362-368`, `InvitationAcceptedData.required` includes `roles`. The PR's stated intent is to remove a silent conformance violation; instead it trades an extra-field violation for a missing-required-field violation.
- **CI red (not introduced by this PR)**: the `RPC Shape Registry Sync` job fails inside the upstream migration `20260506190626_add_role_management_view_ou_implications.sql` with `Postcondition violated: expected 6 implication rows, found 0`. The local supabase container apparently isn't getting the seed permissions applied before that migration runs. Reproducible on `main`. Worth flagging but separate concern; the merge gate will block on it regardless.
- **MINOR**: inline `RoleRef.role_name: string` (non-nullable) in the role-assignment loop is inconsistent with the rest of the type surface (`InvitationDetails.roles[].role_name: string | null`, validate-invitation's `InvitationRoleRef.role_name: string | null`, MockInvitationService).

Migration itself looks good — DROP+CREATE re-issues the `@a4c-rpc-shape: read` tag, defensive backfill before DROP COLUMN, write-site / read-site audit comments are accurate (verified `process_invitation_event` baseline_v4:11064 INSERT does not reference `role`).

_Tee'd locally before posting so the user can see the same notes I'm leaving inline._

lars-tice (commented):
**Re-review at c93eb010 — all three prior findings addressed cleanly. No new blockers.** CI is now fully green (RPC Shape Registry Sync, PL/pgSQL validation, Edge Function ADR check, Deno tests, Temporal Docker build all SUCCESS).

**Fix #1 (BLOCKER, AsyncAPI conformance) — resolved, and stronger than what I suggested.**
The author noticed that emitting the local `roles` array as-is would have shipped entries with `role_id: null`, which violates the AsyncAPI `RoleAssignment.required: [role_id]` constraint at `infrastructure/supabase/contracts/asyncapi/domains/invitation.yaml:255-256`. Resolution is to track a separate `resolvedRoles: Array<{role_id: string; role_name?: string}>` during the role-assignment loop (each entry has the resolved id — pre-supplied or looked up via `api.get_role_by_name`) and emit that on the `invitation.accepted` event. `role_name` is omitted (not emitted-as-null) when absent, matching the schema's non-nullable `type: string` declaration. The comment block above the emit was also rewritten to cite the contract path. This is the schema-correct shape, and it's better than my original suggestion.

**Fix #2 (MEDIUM, type alignment) — resolved cleanly.**
The inline `RoleRef.role_name` is now `string | null` matching the upstream type surface. The new explicit `if (!roleId && !roleName)` guard returns 400 `malformed_role_entry` with an actionable message before hitting `get_role_by_name` — surfaces the actual failure mode ("invitation contained an entry with no identifying info") instead of masking it as `role_lookup_failed: "Cannot find role 'null'"`. The `roleName as string` cast inside the lookup branch is type-narrowed by the preceding guard. Sound.

**Fix #4 (CI red, RPC Shape Registry Sync) — surgical workaround, well-engineered, with the right architectural follow-up identified.**
The phased-apply approach (stage post-baseline migrations aside → start container with baseline only → restore → manual psql apply in three phases with the permissions seed between phase 1 and phase 3) correctly mirrors the dev/prod bootstrap state where `99-seeds/001-permissions-seed.sql` ran post-baseline and before the May 6 postcondition migration. Two observations — neither blockers, one inline:

1. The boundary timestamp `BOUNDARY_TS=20260506012315` is hardcoded. Any future migration whose timestamp falls between that and `20260506190626` and that has its own permission-projection dependency would re-break the workflow. The commit message correctly identifies the long-term fix (promote `99-seeds/001-permissions-seed.sql` to a real migration with `migration repair --status applied` on dev/prod) as a follow-up card.
2. The `resolvedRoles` emit and the new `malformed_role_entry` 400 path are not directly covered by Deno unit tests (the existing `accept-invitation/__tests__/phone-id-resolution.test.ts` only exercises the pure phone-resolution helper). Not a regression — those branches live inside the main `serve` handler and the project doesn't yet have a handler-level integration test pattern for Edge Functions. Worth noting for the next person.

**Migration, handler reference file, validate-invitation, frontend, workflows query-script — unchanged since the prior review and still correct.**

_Tee'd for the user: PR #54 looks good to merge once UAT signs off. The 4-signal verification packet from the original PR description still applies; add the new SQL audit from the e6bda482 commit message — `SELECT (event_data ? 'roles') AS has_roles FROM domain_events WHERE event_type = 'invitation.accepted' ORDER BY created_at DESC LIMIT 5` — must return `has_roles=true` for new acceptances post-deploy._

.github/workflows/rpc-registry-sync.yml
```diff
@@ -42,16 +42,125 @@ jobs:
...
+      # early migrations first, seeding permissions, then applying remaining
+      # migrations.
+      - name: Apply migrations and seed permissions in phased order
         working-directory: infrastructure/supabase
-        run: supabase db reset --local
+        env:
+          PGPASSWORD: postgres
+        run: |
+          PSQL_CMD="psql -h 127.0.0.1 -p 54322 -U postgres -d postgres -v ON_ERROR_STOP=1"
+
```
lars-tice, line 110: **Nit, no action needed for this PR.**

The boundary timestamp `BOUNDARY_TS=20260506012315` is the migration immediately before the postcondition migration (`20260506190626_add_role_management_view_ou_implications.sql`). The phased apply correctly puts the permissions seed between Phase 1 (≤ boundary) and Phase 3 (> boundary), so the postcondition lands in Phase 3 with `permissions_projection` already populated.

Future-proofing concern: if a new migration with timestamp between `20260506012315` and `20260506190626` is ever added (unlikely but possible if someone backdates), and that migration has its own permission-projection dependency, this script will silently revert to the pre-fix failure mode because the new migration would land in Phase 1 (before the seed).

The long-term fix the commit message correctly identifies — promote `infrastructure/supabase/sql/99-seeds/001-permissions-seed.sql` to a real Supabase migration with `migration repair --status applied` on dev/prod — eliminates the boundary-timestamp coupling entirely. Worth seeding as a follow-up card; out of scope for this PR.

infrastructure/supabase/supabase/functions/accept-invitation/index.ts
```diff
@@ -754,6 +783,17 @@ serve(async (req) => {
...
 
+      // Track for the invitation.accepted event_data emit below.
+      // roleId is non-null here (either pre-supplied or just resolved above).
+      // role_name is omitted when null per the AsyncAPI RoleAssignment schema.
+      const resolvedEntry: { role_id: string; role_name?: string } = {
+        role_id: roleId as string,
+      };
+      if (roleName !== null) {
+        resolvedEntry.role_name = roleName;
+      }
```
lars-tice, line 794: Worth noting for future test coverage (not a blocker for this PR): the new `malformed_role_entry` 400 branch (line 744) and the `resolvedRoles` emit shape (line 985) aren't covered by the existing Deno tests at `__tests__/phone-id-resolution.test.ts`, which only exercises the pure phone-resolution helper. Either the loop body needs to be extracted into a pure helper (similar to `resolveInvitationPhonePlaceholder`), or the project needs a handler-level integration test pattern for Edge Functions. Eligible for a follow-up card given the no-deferral rule's "genuinely separate body of work" carve-out — handler-level integration tests are a category the codebase doesn't have yet.

PR #55, opened 2026-05-11
fix(users): rich error surfacing on user-edit form (PR #54 follow-up, UAT S4)

lars-tice:
## Summary

Follow-up to PR #54. UAT Scenario 4 (SUBSET_ONLY_VIOLATION on `modify_user_roles` from the user-edit form) exposed a frontend defect: the form modal showed the bare error code "VALIDATION_FAILED" instead of the actionable violation message. The backend was correct; the frontend dropped violation detail.

Funnels the form's role-modify call through `UsersViewModel.modifyRoles()` (which already has rich-state handling for the page-level `UsersErrorBanner`) instead of calling `commandService.modifyRoles()` directly. Single source of truth for role-modification error surfacing.

## Defect trace

| Path | Handler | Rich rendering? |
|---|---|---|
| Role-only-edit dialog | `UsersViewModel.modifyRoles()` | ✅ wired (page banner with `role-modification-violation` testid) |
| User-edit form (profile+roles) | `UserFormViewModel.submit()` → direct service call | ❌ dropped violations |

## Architect review

Reviewed by `software-architect-dbc` agent (plan at `~/.claude/plans/uat-s4-frontend-error-surfacing-fix.md`). Eight findings — Finding #4 (funnel-through pattern) collapsed Findings #1, #2, #5, and #8 into a single coherent refactor. All addressed:

- **#1**: Refined `submit()` return type to `InviteUserResult | UpdateUserResult | ModifyUserRolesResult` — no anonymous-shape cast.
- **#2**: Single surface — when page banner captures violations/partial, form's `submissionError` is suppressed.
- **#3**: It's not a modal; corrected the plan.
- **#4** (adopted): Funnel through `UsersViewModel.modifyRoles()`.
- **#5**: PARTIAL_FAILURE shape gets the same treatment for free.
- **#6**: Expanded test plan from 2 cases to 6.
- **#7**: Scope discipline OK.
- **#8**: `formatViolationDetails` becomes dead code on role-modify branch (informational, follow-up sweep).

## Changes

| File | Change |
|---|---|
| `frontend/src/viewModels/users/UserFormViewModel.ts` | Inject `usersViewModel` param into `submit()`; funnel role-modify through page VM; suppress `submissionError` when page banner owns the error; prefer `errorDetails.message` over bare `error` in inline fallback |
| `frontend/src/pages/users/UsersManagePage.tsx` | Pass `viewModel` as new 2nd arg to `formViewModel.submit()` |
| `frontend/src/viewModels/users/__tests__/UserFormViewModel.submit.test.tsx` | NEW — 6 Vitest cases including integration render test |

## Test plan

- ✅ Frontend gates: typecheck, lint, build, docs:check all clean.
- ✅ Unit tests: 6/6 pass.
- [ ] Manual re-run of UAT Scenario 4 on dev after deploy:
  - Log in as Lars Tice-Test
  - Open Lars Tice-Test2 in edit panel
  - Try to add provider_admin → Save
  - **Expected**: Page banner appears with heading "Role assignment violation"; inside, `<li data-testid="role-violation-SUBSET_ONLY_VIOLATION">` contains "Role 'provider_admin' has permissions you don't have"; form inline error is gone; no "VALIDATION_FAILED" raw code anywhere
- [ ] Mark UAT Scenario 4 row as Pass in the test plan.

## Verification — 6 test cases

(a) Happy path — profile-only edit (page VM untouched)
(b) Role-violation path (form silent, page VM captures violations)
(c) Partial-failure path (form silent, page VM captures partial)
(d) Non-role failure regression (form surfaces inline, page VM untouched)
(e) Combined profile-success + role-violation (the actual S4 shape)
(f) Integration render — page-VM state drives `UsersErrorBanner` with `role-violation-SUBSET_ONLY_VIOLATION` test-id; bare error code never appears in rendered DOM

## Plan

`~/.claude/plans/uat-s4-frontend-error-surfacing-fix.md` (revised twice after architect review).

### DIFF
--- a/frontend/src/viewModels/users/UserFormViewModel.ts
+++ b/frontend/src/viewModels/users/UserFormViewModel.ts
@@ -32,12 +32,13 @@ import type {
   RoleReference,
   InviteUserResult,
   UpdateUserResult,
-  UserVoidResult,
+  ModifyUserRolesResult,
   EmailLookupResult,
   NotificationPreferences,
   InvitationPhone,
   UserListItem,
 } from '@/types/user.types';
+import type { UsersViewModel } from './UsersViewModel';
 import {
   validateEmail,
   validateFirstName,
@@ -893,20 +894,24 @@ export class UserFormViewModel {
    * Submit the form
    *
    * In create mode: Sends invitation via commandService.inviteUser()
-   * In edit mode: Updates user profile via commandService.updateUser()
+   * In edit mode: Updates user profile via commandService.updateUser(), then
+   * (if profile update succeeded and roles changed) delegates role modification
+   * to `usersViewModel.modifyRoles(...)` — the page-level handler that owns
+   * `lastRoleViolations` / `lastRolePartialFailure` state for the rich
+   * `UsersErrorBanner` rendering. Going through the page VM (instead of calling
+   * `commandService.modifyRoles` directly) is the single source of truth for
+   * role-modification error surfacing: violation/partial detail flows to the
+   * page banner, the form's `submissionError` is suppressed when the banner
+   * owns it.
    *
-   * Return union intentionally contains UserVoidResult: on edit mode, if
-   * profile update succeeds but subsequent modifyRoles() fails, `result` is
-   * reassigned to the modifyRoles UserVoidResult so consumers see the
-   * role-modification error (see the `result = roleResult` reassignment
-   * around line 969 below). Dropping UserVoidResult here would force a
-   * later consumer to either cast or narrow via success-property access.
-   * PR #32 review item 6 proposed removing this arm as "phantom"; it is
-   * structurally necessary.
+   * Return union: on edit mode with role changes, `result` is reassigned to
+   * the modifyRoles result when role modification fails, so callers see the
+   * role error (rich `violations` / `partial` fields on `ModifyUserRolesResult`).
    */
   async submit(
-    commandService: IUserCommandService
-  ): Promise<InviteUserResult | UpdateUserResult | UserVoidResult> {
+    commandService: IUserCommandService,
+    usersViewModel: UsersViewModel
+  ): Promise<InviteUserResult | UpdateUserResult | ModifyUserRolesResult> {
     // Touch all fields to show validation errors
     this.touchAllFields();
 
@@ -945,7 +950,7 @@ export class UserFormViewModel {
     });
 
     try {
-      let result: InviteUserResult | UpdateUserResult | UserVoidResult;
+      let result: InviteUserResult | UpdateUserResult | ModifyUserRolesResult;
 
       if (this.mode === 'create') {
         // Create mode: Send invitation
@@ -962,33 +967,35 @@ export class UserFormViewModel {
         log.debug('Updating user profile', { userId: updateRequest.userId });
         result = await commandService.updateUser(updateRequest);
 
-        // Handle role changes (if profile update succeeded and roles changed)
+        // Handle role changes (if profile update succeeded and roles changed).
+        // Delegate to the page VM's modifyRoles handler — it captures
+        // `lastRoleViolations` and `lastRolePartialFailure` on itself so the
+        // page-level UsersErrorBanner can render the rich violation list
+        // (`data-testid="role-modification-violation"`) or partial-failure
+        // recovery banner (`data-testid="role-modification-partial-warning"`).
+        // The page VM also logs structured warnings for both failure modes.
         if (result.success && this.hasRoleChanges) {
           log.debug('Processing role changes', {
             userId: this.editingUserId,
             rolesToAdd: this.rolesToAdd,
             rolesToRemove: this.rolesToRemove,
           });
-          const roleResult = await commandService.modifyRoles({
+          const roleResult = await usersViewModel.modifyRoles({
             userId: this.editingUserId!,
             roleIdsToAdd: this.rolesToAdd,
             roleIdsToRemove: this.rolesToRemove,
           });
 
           if (!roleResult.success) {
-            // Role modification failed - return role error instead
+            // Role modification failed — return role error so caller can react.
+            // The page VM has already populated `lastRoleViolations` /
+            // `lastRolePartialFailure` for the banner; the failure branch below
+            // suppresses the form's inline submissionError to avoid duplicate
+            // error surfaces.
             result = roleResult;
-            log.warn('Role modification failed', {
-              userId: this.editingUserId,
-              error: roleResult.error,
-            });
-          } else {
-            log.info('Role modification successful', {
-              userId: this.editingUserId,
-              added: this.rolesToAdd.length,
-              removed: this.rolesToRemove.length,
-            });
           }
+          // Success path: page VM has already set `successMessage = 'Roles
+          // updated'`; no additional log here (the page VM logs internally).
         }
       }
 
@@ -1002,22 +1009,40 @@ export class UserFormViewModel {
             log.info('User profile updated successfully', { userId: this.editingUserId });
           }
         } else {
-          this.submissionError = result.error ?? 'An error occurred';
-          // Extract detailed error info for display
-          // The context contains the full errorDetails from the Edge Function
-          const ctx = result.errorDetails?.context as Record<string, unknown> | undefined;
-          if (ctx) {
-            this.submissionErrorDetails = {
-              code: result.errorDetails?.code ?? (ctx.code as string),
-              details: this.formatViolationDetails(ctx),
-            };
-          } else {
+          // If the page VM has captured structured role-modification state
+          // (violations or partial-failure), that surface owns the error
+          // display. Suppress the form's inline submissionError to avoid
+          // rendering the same error twice (architect Finding #2).
+          const handledByPageBanner =
+            usersViewModel.lastRoleViolations !== null ||
+            usersViewModel.lastRolePartialFailure !== null;
+
+          if (handledByPageBanner) {
+            this.submissionError = null;
             this.submissionErrorDetails = null;
+          } else {
+            // Non-role failures (profile-update error, network error, etc.)
+            // still surface inline in the form. Prefer the rich
+            // `errorDetails.message` over the bare `error` code string.
+            this.submissionError =
+              result.errorDetails?.message ?? result.error ?? 'An error occurred';
+            // Extract detailed error info for display.
+            // The context contains the full errorDetails from the Edge Function.
+            const ctx = result.errorDetails?.context as Record<string, unknown> | undefined;
+            if (ctx) {
+              this.submissionErrorDetails = {
+                code: result.errorDetails?.code ?? (ctx.code as string),
+                details: this.formatViolationDetails(ctx),
+              };
+            } else {
+              this.submissionErrorDetails = null;
+            }
           }
           log.warn('Form submission failed', {
             mode: this.mode,
             error: result.error,
             errorDetails: result.errorDetails,
+            handledByPageBanner,
           });
         }
       });
--- a/frontend/src/viewModels/users/UserFormViewModel.ts
+++ b/frontend/src/viewModels/users/UserFormViewModel.ts
@@ -949,6 +949,16 @@ export class UserFormViewModel {
       this.submissionError = null;
     });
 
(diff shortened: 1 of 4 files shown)


frontend/src/viewModels/users/UserFormViewModel.ts
```diff
@@ -1002,22 +1009,40 @@ export class UserFormViewModel {
...
-              details: this.formatViolationDetails(ctx),
-            };
-          } else {
+          // If the page VM has captured structured role-modification state
+          // (violations or partial-failure), that surface owns the error
+          // display. Suppress the form's inline submissionError to avoid
+          // rendering the same error twice (architect Finding #2).
+          const handledByPageBanner =
+            usersViewModel.lastRoleViolations !== null ||
+            usersViewModel.lastRolePartialFailure !== null;
```
lars-tice: **MEDIUM — stale-state suppression bug.**

The suppression condition reads the page VM's sticky state with no provenance check: it doesn't distinguish "this submit's `modifyRoles` call populated these fields" from "a prior submit populated these fields and nobody has reset them since." The reset points for `lastRoleViolations` / `lastRolePartialFailure` are only `UsersViewModel.modifyRoles()` (line 1189-1190 in that file) and `clearError()`. Neither fires on the form-submit start path.

Concrete failure scenario:
1. User triggers a role change → Save → `modifyRoles` returns `SUBSET_ONLY_VIOLATION` → `pageVm.lastRoleViolations` populated.
2. User reverts the role change (so `hasRoleChanges = false`) but updates the first name → Save.
3. `commandService.updateUser()` fails (network error, processing_error, anything non-role).
4. `hasRoleChanges = false`, so the `modifyRoles` branch is skipped — `pageVm.lastRoleViolations` retains its value from step 1.
5. Failure branch runs: `handledByPageBanner = (lastRoleViolations !== null || lastRolePartialFailure !== null) === true` → form's `submissionError` is suppressed.
6. UI: page banner still shows the stale violation; the new network/processing error is hidden.

Same shape for `lastRolePartialFailure`. Same shape when `hasRoleChanges = true` AND `updateUser` fails (so `modifyRoles` is never called this submit).

Cleanest local fix — snapshot-before-compare (no change to page-VM semantics, doesn't dismiss a banner the user might be intentionally reading):

```ts
// Just before the try block, after `this.isSubmitting = true`:
const initialRoleViolations = usersViewModel.lastRoleViolations;
const initialRolePartialFailure = usersViewModel.lastRolePartialFailure;

// In the suppression check:
const handledByPageBanner =
  usersViewModel.lastRoleViolations !== initialRoleViolations ||
  usersViewModel.lastRolePartialFailure !== initialRolePartialFailure;
```

Reference-identity comparison is what you want here — `modifyRoles` always assigns a fresh array/object on populate (lines 1205, 1216 of UsersViewModel), so reference equality only holds when this submit didn't touch them. Equivalent boolean alternative if you'd rather track explicit intent: a local `let invokedRoleModify = false` set inside the role-modify branch and used to gate the suppression — slightly less subtle, slightly more bookkeeping.

The other approach is to call `usersViewModel.clearError()` at submit start. Functionally correct but changes the semantics of banner lifetime (it currently sticks until the user dismisses or runs another role-modify, which is reasonable). I'd prefer snapshot/compare to keep banner semantics where they are.

frontend/src/viewModels/users/__tests__/UserFormViewModel.submit.test.tsx
```diff
@@ -0,0 +1,343 @@
...
+    expect(vm.submissionError).toBe(null);
+
+    // Page VM captured the rich state for UsersErrorBanner to render
+    expect(pageVm.lastRoleViolations).toEqual(violations);
+    expect(pageVm.lastRoleViolations?.[0].error_code).toBe('SUBSET_ONLY_VIOLATION');
+    expect(pageVm.lastRoleViolations?.[0].message).toBe(
+      'Role "provider_admin" has permissions you don\'t have'
+    );
+  });
+});
```
lars-tice, line 435: **Add a stale-state test case (g) to cover the bug noted on `UserFormViewModel.ts:1018`.**

Proposed:

```ts
it('(g) stale page-VM state from a prior submit does not suppress a fresh non-role failure', async () => {
  const vm = new UserFormViewModel(mockAssignableRoles, 'edit', mockExistingUser);
  vm.toggleRole(PROVIDER_ADMIN_ROLE_ID);

  // Submit #1: role-violation populates pageVm.lastRoleViolations
  (cs.updateUser as ReturnType<typeof vi.fn>).mockResolvedValue({ success: true } satisfies UpdateUserResult);
  const violations: RoleAssignmentViolation[] = [{
    role_id: PROVIDER_ADMIN_ROLE_ID, role_name: 'provider_admin',
    error_code: 'SUBSET_ONLY_VIOLATION',
    message: 'Role "provider_admin" has permissions you don\'t have',
  }];
  let nextRoleResult: ModifyUserRolesResult = {
    success: false, error: 'VALIDATION_FAILED', violations,
    errorDetails: { code: 'VALIDATION_FAILED', message: violations[0].message },
  };
  const pageVm = makePageViewModel(async () => nextRoleResult);
  await vm.submit(cs, pageVm);
  expect(pageVm.lastRoleViolations).toEqual(violations);  // sanity: state captured

  // User reverts role change → no role changes for submit #2
  vm.toggleRole(PROVIDER_ADMIN_ROLE_ID);
  expect(vm.hasRoleChanges).toBe(false);

  // Submit #2: profile update fails with a non-role error
  (cs.updateUser as ReturnType<typeof vi.fn>).mockResolvedValue({
    success: false, error: 'PROCESSING_ERROR',
    errorDetails: { code: 'PROCESSING_ERROR', message: 'Connection refused' },
  } satisfies UpdateUserResult);

  await vm.submit(cs, pageVm);

  // The new non-role error MUST surface inline (the page banner's stale state
  // is not what failed this time).
  expect(vm.submissionError).toBe('Connection refused');
  // pageVm.modifyRoles must not have been called on submit #2.
  expect((pageVm.modifyRoles as ReturnType<typeof vi.fn>).mock.calls.length).toBe(1);
});
```

With the current code this test fails (the form's `submissionError` is null because stale `lastRoleViolations` falsely triggers suppression). With the snapshot/compare fix it passes.

lars-tice (commented):
Funnel-through approach is the right call and the test coverage is genuinely good (6 cases including a real-DOM render test against UsersErrorBanner with stable test-ids). The architect-review collapse of findings #1/#2/#5/#8 into the single funnel-through refactor was the right design move.

**One real defect found — stale-state suppression.** See inline on UserFormViewModel.ts:1018. The suppression check reads `usersViewModel.lastRoleViolations !== null || usersViewModel.lastRolePartialFailure !== null`, but those fields are only ever reset at the start of `UsersViewModel.modifyRoles()` (line 1189-1190) or via explicit `clearError()`. On a follow-up submit where the form goes through a path that does NOT call `modifyRoles` (most concretely: a previous submit captured a violation; user reverts the role change; user changes profile fields; updateUser fails), the stale state from the prior submit is still set → `handledByPageBanner` evaluates true → the new profile-update error is suppressed and the user sees the stale banner instead of the actual fresh error. Easy local fix proposed inline.

**Missing test (g) for the above — see inline on the test file.** The 6 cases all start from a clean page-VM and don't exercise sticky state across submits.

**Minor robustness nit on test assertions** — the existing case-(b) suppression check passes both when the suppression logic is correct and when the suppression logic is permanently broken (form's `submissionError` happens to be null for other reasons). Worth strengthening with a negative-control assertion. Inline.

CI hasn't been triggered yet on this branch — no `statusCheckRollup` rows from gh. Worth pushing an empty change or asking GitHub to re-run on this head to confirm rpc-registry-sync and frontend gates pass before merge. Otherwise the local gates listed in the PR description (typecheck, lint, build, docs:check, 6/6 vitest) all came back clean, so I don't expect surprises.

Migration, edge functions, workflows, and database.types — untouched by this PR, no concerns.

_Tee'd for the user: one MEDIUM defect to address before merge, two MINORs that can land in-PR per the no-deferral-to-cards rule (small, in scope, would not naturally be a separate body of work)._

frontend/src/viewModels/users/__tests__/UserFormViewModel.submit.test.tsx
```diff
@@ -0,0 +1,343 @@
...
+
+    expect(result.success).toBe(false);
+    // The role-failure shape replaces the profile result
+    expect((result as ModifyUserRolesResult).violations).toEqual(violations);
+    // Form's inline error is suppressed — page banner owns it
+    expect(vm.submissionError).toBe(null);
+    expect(vm.submissionErrorDetails).toBe(null);
+    // Page VM captured the violation
+    expect(pageVm.lastRoleViolations).toEqual(violations);
+  });
```
lars-tice, line 180: **Nit — strengthen the causal assertion.**

The `expect(vm.submissionError).toBe(null)` at line 176 passes both when the suppression logic is firing correctly AND when the form is broken in a way that leaves `submissionError` null for unrelated reasons. Worth a negative-control variant that does NOT populate the page VM state on the violation path and asserts the inline error appears — that way both directions of the conditional are exercised. Drop-in next to case (b):

```ts
it('(b\') role-violation but page VM fails to capture state → form surfaces inline', async () => {
  const vm = new UserFormViewModel(mockAssignableRoles, 'edit', mockExistingUser);
  vm.toggleRole(PROVIDER_ADMIN_ROLE_ID);
  (cs.updateUser as ReturnType<typeof vi.fn>).mockResolvedValue({ success: true } satisfies UpdateUserResult);

  // Page VM returns the violation result but DOESN'T populate lastRoleViolations
  // (simulating a regression in UsersViewModel.modifyRoles).
  const pageVm = {
    lastRoleViolations: null,
    lastRolePartialFailure: null,
    modifyRoles: vi.fn(async () => ({
      success: false, error: 'VALIDATION_FAILED',
      violations: [{ role_id: '...', role_name: 'x', error_code: 'SUBSET_ONLY_VIOLATION', message: 'msg' }],
      errorDetails: { code: 'VALIDATION_FAILED', message: 'msg' },
    })),
  } as unknown as UsersViewModel;

  await vm.submit(cs, pageVm);

  // Without the page-VM capture, the suppression doesn't fire and the form falls
  // through to inline display.
  expect(vm.submissionError).not.toBe(null);
});
```

Low priority — the integration render test (f) effectively covers the success path of the same wiring. Optional.

lars-tice:
**Addressed all three findings from the prior review in commit `83d05ed7`.**

### MEDIUM — stale-state suppression (`UserFormViewModel.ts:1018`)

Switched from a direct null-check to a snapshot-before-compare at submit start:

```ts
const initialRoleViolations = usersViewModel.lastRoleViolations;
const initialRolePartialFailure = usersViewModel.lastRolePartialFailure;
// ...
const handledByPageBanner =
  usersViewModel.lastRoleViolations !== initialRoleViolations ||
  usersViewModel.lastRolePartialFailure !== initialRolePartialFailure;
```

Reference comparison is sound because `UsersViewModel.modifyRoles` always assigns a fresh array/object when populating (lines 1205, 1216). Did NOT use `clearError()` at submit start — that would change banner-lifetime semantics for the legitimate "user reads violation between failed submit and retry" path.

### MINOR — test `(g)` for the stale-state scenario

New test added immediately before case `(e)`. Sequence: submit #1 captures a violation; user reverts the role change; submit #2 has no role changes and `updateUser` fails with a non-role error. Asserts the new error surfaces inline despite the page VM's sticky `lastRoleViolations`. Test **fails** against the prior `!== null` check and **passes** against the snapshot/compare fix — exactly the regression guard the review proposed.

### MINOR — test `(b')` negative control

New test added immediately before `(g)`. Page VM returns a violation result but leaves its own `lastRoleViolations` null (simulates a regression in `UsersViewModel.modifyRoles`). Asserts suppression does NOT fire and `vm.submissionError` is non-null. Gives case `(b)` a causal counterpart so the suppression assertion can't pass for unrelated reasons.

### Local gates (re-run against the new commit)

| Gate | Result |
|---|---|
| typecheck | ✓ |
| lint (`--max-warnings 0`) | ✓ |
| docs:check | ✓ (0 issues) |
| build | ✓ |
| vitest | ✓ 8/8 (6 original + `(b')` + `(g)`) |

CI checks on `83d05ed7` should run automatically.

PR #59, opened 2026-05-12
refactor(services): PR-D observability backfill — logIfPostgrestError retrofit

lars-tice:
## Summary

**PR-D** of the `migrate-services-to-api-rpc-envelope` card. Closes the observability regression pattern surfaced by PR #58 architect review F1: `throwIfPostgrestError` closed the PII defense surface but not the observability surface, and the return-contract path needs its own SDK-boundary helper. `logIfPostgrestError` was promoted to envelope.ts in PR #58 (commit `3a616414`) and applied to ClientService's 4 lifecycle methods. PR-D applies it to the other **14 return-contract regression sites** across 4 services where the same pre-migration `log.error('Failed to <verb>', { error })` was silently dropped during the PR-A/B migrations.

**Card**: `dev/active/migrate-services-to-api-rpc-envelope/`
**Plan**: `~/.claude/plans/ddoes-it-make-sense-lucky-dongarra.md`
**Architect verdict**: APPROVE WITH PLAN REVISION (Q2 → verb-from-rpcName, applied)
**Prior PRs**: #56 (PR-A), #57 (PR-B), #58 (PR-C — all merged)

## Scope clarifier (per PR-D architect review NT-1)

PR-D fixes the **14 regression sites** where pre-migration `log.error` on PostgREST failure was silently dropped during PR-A/B. It does NOT touch the broader "centralize all PostgREST-error logs at the SDK boundary" invariant:

- `SupabaseUserCommandService` retains **~5 return-contract sites** with bespoke inline `if (env.postgrestError) log.error('...')` blocks. These were **preserved intentionally** during PR-A/B with custom message strings — not regressions. Moving them to the SDK-boundary helper would flatten the bespoke verb strings and is out of PR-D scope.
- `SupabaseRoleService.createRole` retains the `[DIAG:createRole:RPC_ERROR]` block per PR #56 NT-1 (richer fields than the helper). Out of PR-D scope.

After PR-D merges, the **F1 regression class is closed**: no return-contract envelope method silently absorbs a PostgREST failure. The migration card's close-out language should reflect "regression class closed" rather than "all PostgREST logs centralized."

## Retrofits (14 sites, 4 services)

| Service | Methods | Verbs |
|---|---|---|
| `OrgCommandService` | update/deactivate/reactivate/deleteOrganization | `'update organization'`, etc. |
| `OrgUnitService` | create/update/deactivate/reactivate/deleteUnit | `'create organization unit'`, etc. |
[...]

PR #58, opened 2026-05-11
refactor(services): PR-C closeout — migrate Clients/OrgQuery/Subdomain + activate ESLint rule

lars-tice:
## Summary

**PR-C of 3** — final wave of the apiRpc/apiRpcEnvelope migration card. Migrates the remaining 3 service files and **activates the `no-restricted-syntax` ESLint rule** that forbids new direct `.schema('api').rpc(...)` callers. Closes out the 3-PR plan.

**Card**: `dev/active/migrate-services-to-api-rpc-envelope/`
**Plan**: `~/.claude/plans/ddoes-it-make-sense-lucky-dongarra.md` (v2)
**Prior PRs**: #56 (PR-A, merged) + #57 (PR-B, merged)

## Migrations in this PR

| File | Sites |
|---|---|
| `clients/SupabaseClientService.ts` | 25 envelope |
| `organization/SupabaseOrganizationQueryService.ts` | 4 read + 1 envelope = 5 sites |
| `organization/getOrganizationSubdomainInfo.ts` | 1 read (Q5 refactor) |
| `eslint.config.js:124-131` | placeholder → active `no-restricted-syntax` rule |
| **TOTAL** | **31 sites + rule** |

**+1 site vs. plan estimate of 30** — `getOrganizationDetails` was misclassified as read in the inventory; M3 registry tags it as envelope. Migrated via `apiRpcEnvelope`.

## F3 pre-PR-C inventory refresh (architect-required)

Per architect F3 from PR #57 review, re-ran inventory with multi-line-aware grep before opening this PR:

```bash
grep -lE "\.schema\(['\"]api['\"]\)" frontend/src/services/ -r | grep -v __tests__ | grep -v CLAUDE.md
```

Returns exactly 4 files:
1. `auth/supabase.service.ts` — SDK helper (allow-listed, must not migrate)
2. `organization/SupabaseOrganizationQueryService.ts` — migrated in this PR
3. `organization/getOrganizationSubdomainInfo.ts` — migrated in this PR
4. `clients/SupabaseClientService.ts` — migrated in this PR

**Zero new direct callers landed in the PR-A→PR-C window.** Q4 (accept the window) was sound.

## Q5 — `.single<T>()` refactor in `getOrganizationSubdomainInfo.ts`

[...]

PR #60, opened 2026-05-12
feat(manage-user): retrofit Pattern A v2 read-back to deactivate operation

lars-tice:
## Summary

Closes the F1-class silent-failure gap on the Edge Function deactivate path. Pre-retrofit, the function emitted `user.deactivated` and returned `{success: true}` unconditionally after the emit RPC succeeded. If the handler raised mid-update, the BEFORE INSERT trigger persisted `processing_error` on the event row WITHOUT rolling back, and the function returned success while `users.is_active` stayed true.

This is the same regression class PR #58/#59 closed at the frontend SDK boundary — closed here at the Edge Function tier. **First Edge Function adopter of Pattern A v2.**

**Card**: `dev/active/manage-user-deactivate-pattern-a-v2-retrofit/`
**Plan**: `~/.claude/plans/ddoes-it-make-sense-lucky-dongarra.md`
**Architect verdict**: APPROVE WITH PLAN REVISION (Q1 + Q2 + Q3 + Q6 absorbed inline)

## New shared helper

`infrastructure/supabase/supabase/functions/_shared/rpc-readback.ts`:

- `checkProjectionReadback<T>(client, eventId, table, lookupKey, lookupValue, expectedState)`
- **Architect Q2**: `expectedState: Record<string, unknown>` predicate — non-negotiable. Without it the helper can't distinguish "handler updated correctly" from "handler threw before update, pre-existing stale row visible."
- **Architect Q6**: `maskPii` on `processing_error` before concatenation — future-proofs against Rule 16 regressions across the next ~5-10 retrofits.
- **Architect Q1**: transactional-model docblock — wire-tier port is *safer* than SQL RPC's intra-transaction read-back, not less safe (no MVCC snapshot ambiguity).
- Designed for byte-equivalent reuse by the sibling `manage-user-reactivate-pattern-a-v2-retrofit` card with `expectedState: { is_active: true }`.

## Edge Function modification

`manage-user/index.ts`:
- `DEPLOY_VERSION` bumped from `v15-modify-roles-extracted` to `v16-deactivate-pattern-a-v2-readback`
- Capture `eventId` from emit RPC (previously discarded)
- Call `checkProjectionReadback` with `{is_active: false}` on **deactivate path only**
- `reactivate` path intentionally NOT retrofitted — sibling card next
- Surface `eventId` additively on `ManageUserResponse`

## Tests (8 Deno cases — architect Q3 expansion)

`manage-user/__tests__/deactivate-readback.test.ts`:

| # | Case |
|---|---|
| 1 | Happy path: emit + projection updated + no processing_error → success |
| 2 | Handler-throws: row missing + processing_error set → masked error |
| 3 | Race-safe: projection looks updated but processing_error set → masked error |
| 4 | Expected-state mismatch: row exists but doesn't match predicate |
| 5 | **NEW (architect Q3)**: helper queries `domain_events` by captured event_id (NOT stream_id) |
| 6 | **NEW (architect Q3)**: helper does NOT swallow `domain_events` query errors |
| 7 | **NEW (architect Q3)**: helper does NOT swallow projection-table query errors |
| 8 | **NEW (architect Q6)**: PII masking on processing_error string |

All 8 pass locally.

## Frontend additive (architect Q4)

- `eventId?: string` added to `UserRpcEnvelope` (`frontend/src/types/user.types.ts`) AND `ClientRpcEnvelope` (`frontend/src/types/client.types.ts`) for cross-service uniformity per architect's recommendation. The audit-link capability is now uniformly available on every envelope-returning service call going forward.
- `SupabaseUserCommandService.deactivateUser` plumbs `eventId` through on the success path additively. No breaking change to `UserVoidResult` or other return types.

## Auth-ban failure semantics (architect Q3, out of scope per plan)

Pre-existing behavior preserved verbatim: `auth.admin.updateUserById` ban failure logs a warning + returns `{success: true}` to the frontend. This is a pre-existing design — a deactivated user could still log in if the auth ban call fails. Not introduced by this retrofit; mentioned here for reviewer clarity. Out of scope per the architect-approved plan.

## Test plan

- [x] `cd frontend && npm run typecheck` — green
- [x] `cd frontend && npm run lint -- --max-warnings 0` — green
- [x] `cd frontend && npm run test -- --run` — 576 pass / 56 fail (pre-existing baseline unchanged from PR #59 merge). Zero regressions.
- [x] `cd frontend && npm run build` — green
- [x] `cd infrastructure/supabase/supabase/functions && deno test --allow-net manage-user/__tests__/` — 8/8 pass
- [ ] Manual smoke (post-deploy): force a handler failure (e.g., temporarily revoke handler's GRANT on `users`), confirm `{success: false, error: 'Event processing failed: ...'}` surfaces in the UI

## Card follow-up

After this PR merges, seed `dev/active/manage-user-reactivate-pattern-a-v2-retrofit/` as the sibling card. Reuses `_shared/rpc-readback.ts` byte-equivalently with `expectedState: { is_active: true }`. ~30 minutes of work.

### DIFF
--- a/frontend/src/types/client.types.ts
+++ b/frontend/src/types/client.types.ts
@@ -707,11 +707,12 @@ export interface ClientRpcEnvelope {
   success: boolean;
   error?: string;
   /**
-   * Captured event_id from the underlying `emit_domain_event` call. Additive
-   * on success responses for audit-log deep-linking. Optional — consumers
-   * are not required to surface it. First introduced 2026-05-12 alongside
-   * the `manage-user.deactivate` Pattern A v2 read-back retrofit; available
-   * uniformly on every envelope-returning service call going forward.
+   * Captured event_id from the underlying `emit_domain_event` call. **Reserved
+   * for future use** — no `api.*` client RPC currently emits this field on
+   * its success envelope. Added 2026-05-12 alongside `UserRpcEnvelope.eventId`
+   * for cross-service uniformity; will be populated as individual client
+   * write RPCs adopt the audit-deep-linking pattern. Optional throughout —
+   * consumers are not required to surface it.
    */
   eventId?: string;
 }
--- a/frontend/src/types/client.types.ts
+++ b/frontend/src/types/client.types.ts
@@ -706,6 +706,14 @@ export type ClientProjectionRow = Omit<
 export interface ClientRpcEnvelope {
   success: boolean;
   error?: string;
+  /**
+   * Captured event_id from the underlying `emit_domain_event` call. Additive
+   * on success responses for audit-log deep-linking. Optional — consumers
+   * are not required to surface it. First introduced 2026-05-12 alongside
+   * the `manage-user.deactivate` Pattern A v2 read-back retrofit; available
+   * uniformly on every envelope-returning service call going forward.
+   */
+  eventId?: string;
 }
 
 /** Response for api.register_client / api.update_client / api.admit_client / api.discharge_client */
--- a/infrastructure/supabase/supabase/functions/_shared/rpc-readback.ts
+++ b/infrastructure/supabase/supabase/functions/_shared/rpc-readback.ts
@@ -75,7 +75,10 @@ export type ProjectionReadbackResult<T> =
  * @param table - Projection table to read back (e.g. `'users'`,
  *   `'roles_projection'`).
  * @param lookupKey - Primary-key column on the projection (typically `'id'`).
- * @param lookupValue - Value to filter the lookupKey on.
+ * @param lookupValue - Value to filter the lookupKey on. Typed `string | number`
+ *   to cover both UUID-keyed (deactivate, reactivate, role, OU) and surrogate-
+ *   integer-keyed projections without forcing every caller to coerce. `.eq()`
+ *   accepts both at runtime.
  * @param expectedState - Object of column→value pairs conjoined into the
  *   read-back query via `.eq()` chains. This is load-bearing: without it,
  *   `IF NOT FOUND` cannot distinguish "handler updated correctly" from
@@ -94,7 +97,7 @@ export async function checkProjectionReadback<T>(
   eventId: string,
   table: string,
   lookupKey: string,
-  lookupValue: string,
+  lookupValue: string | number,
   expectedState: Record<string, unknown>,
 ): Promise<ProjectionReadbackResult<T>> {
   // Check 1: read back with expected-state predicate.
--- a/infrastructure/supabase/supabase/functions/_shared/rpc-readback.ts
+++ b/infrastructure/supabase/supabase/functions/_shared/rpc-readback.ts
@@ -1,0 +1,158 @@
+/**
+ * Pattern A v2 read-back helper for Edge Functions.
+ *
+ * Verifies that a `domain_events` emit produced the expected projection state
+ * AND captures the race-safe `processing_error` on the captured event_id.
+ *
+ * # Two-check contract
+ *
+ * 1. **`IF NOT FOUND` on the projection read-back** — catches the case where
+ *    the row is genuinely missing OR the expected-state predicate doesn't
+ *    match (e.g., handler ran but didn't update; handler threw before update).
+ * 2. **`processing_error` check on the captured event_id** — catches the
+ *    handler-update-then-throw window AND the case where the projection
+ *    looked updated for an unrelated reason.
+ *
+ * # Transactional model — SQL RPC vs Edge Function
+ *
+ * In the SQL precedent (`api.delete_user`, `api.update_role`, etc.) the
+ * `api.emit_domain_event(...)` call + projection read-back + `processing_error`
+ * SELECT all execute inside a single transaction. The `BEFORE INSERT` trigger
+ * fires inside that same transaction, persists `processing_error` to the
+ * in-flight `NEW` row, and the subsequent SELECT sees it via PG's self-write
+ * snapshot visibility.
+ *
+ * In the Edge Function port (this helper), `supabaseAdmin.rpc('emit_domain_event', ...)`
+ * is one wire round-trip / one transaction; the read-back queries are separate
+ * transactions over the wire. After the emit RPC commits, the `domain_events`
+ * row (including `processing_error`) is durably visible — no MVCC snapshot
+ * ambiguity. The wire-tier port is *safer*, not less safe.
+ *
+ * # PII masking
+ *
+ * `processing_error` strings are concatenated into the response body via
+ * `maskPii` (byte-equivalent to frontend masker). Handler `RAISE EXCEPTION`
+ * strings MUST NOT interpolate identifiers (Rule 16 in
+ * `infrastructure-guidelines/SKILL.md`); this helper is the boundary that
+ * future-proofs against handler-side regressions.
+ *
+ * # Reference
+ *
+ * - ADR: `documentation/architecture/decisions/adr-rpc-readback-pattern.md`
+ *   §"Pattern A v2 (Resolved)"
+ * - SQL precedent: `api.delete_user` at
+ *   `infrastructure/supabase/supabase/migrations/20260427205333_extract_delete_user_rpc.sql:107-133`
+ * - First adopter: `manage-user.deactivate` (this PR)
+ */
+
+import type { SupabaseClient } from 'https://esm.sh/@supabase/supabase-js@2';
+import { maskPii } from './maskPii.ts';
+
+/**
+ * Loose type alias matching the AnySchemaSupabaseClient pattern used elsewhere
+ * in `_shared` and the frontend SDK boundary. Required because `supabaseAdmin`
+ * is configured with the default 'public' schema but reads from `api` and
+ * other tables.
+ */
+// deno-lint-ignore no-explicit-any
+type AnyClient = SupabaseClient<any, any, any>;
+
+export type ProjectionReadbackResult<T> =
+  | { success: true; row: T }
+  | { success: false; error: string };
+
+/**
+ * Performs the Pattern A v2 two-check read-back contract for an Edge Function
+ * that has just emitted a domain event and needs to verify the projection
+ * updated as expected.
+ *
+ * @param client - Supabase service-role client (must have read access to the
+ *   projection table AND `domain_events`).
+ * @param eventId - The captured event_id from `emit_domain_event` return.
+ *   This is the race-safe lookup key; never query `domain_events` by
+ *   `stream_id` or "latest event" — concurrent emitters on the same stream
+ *   would produce false negatives.
+ * @param table - Projection table to read back (e.g. `'users'`,
+ *   `'roles_projection'`).
+ * @param lookupKey - Primary-key column on the projection (typically `'id'`).
+ * @param lookupValue - Value to filter the lookupKey on.
+ * @param expectedState - Object of column→value pairs conjoined into the
+ *   read-back query via `.eq()` chains. This is load-bearing: without it,
+ *   `IF NOT FOUND` cannot distinguish "handler updated correctly" from
+ *   "handler threw before update, pre-existing stale row visible." For
+ *   deactivate: `{ is_active: false }`. For reactivate: `{ is_active: true }`.
+ *   For update_role-style multi-column updates: pass all columns.
+ * @returns `{success: true, row: T}` on the happy path; `{success: false,
+ *   error: 'Event processing failed: <masked>'}` on either failure check.
+ *
+ * The error string is structured for the frontend to surface verbatim in
+ * toast/banner UIs — the boundary helper does the masking so no caller is
+ * responsible for it.
+ */
+export async function checkProjectionReadback<T>(
+  client: AnyClient,
+  eventId: string,
+  table: string,
+  lookupKey: string,
+  lookupValue: string,
+  expectedState: Record<string, unknown>,
+): Promise<ProjectionReadbackResult<T>> {
+  // Check 1: read back with expected-state predicate.
+  // If NOT FOUND: query `processing_error` on the captured event_id and
+  // surface the masked failure reason. Surface query errors verbatim
+  // (RLS denials, missing GRANTs would be silent-failure modes if swallowed).
+  let readBackQuery = client.from(table).select('*').eq(lookupKey, lookupValue);
+  for (const [column, value] of Object.entries(expectedState)) {
+    readBackQuery = readBackQuery.eq(column, value);
+  }
+  const readBackResult = await readBackQuery.maybeSingle();
+  if (readBackResult.error) {
+    return {
+      success: false,
+      error: `Read-back query failed: ${maskPii(readBackResult.error.message)}`,
+    };
+  }
+
+  if (!readBackResult.data) {
+    const procErrorResult = await client
+      .from('domain_events')
+      .select('processing_error')
+      .eq('id', eventId)
+      .maybeSingle();
+    if (procErrorResult.error) {
+      return {
+        success: false,
+        error: `processing_error lookup failed: ${maskPii(procErrorResult.error.message)}`,
+      };
+    }
+    const procError = procErrorResult.data?.processing_error as string | null | undefined;
+    return {
+      success: false,
+      error: `Event processing failed: ${maskPii(procError ?? 'projection read-back returned no row')}`,
+    };
+  }
+
+  // Check 2: race-safe `processing_error` check on captured event_id, even
+  // when the read-back looks fine. Catches the partial-update-then-throw
+  // window where the handler successfully wrote some state before raising.
+  const procErrorResult = await client
+    .from('domain_events')
+    .select('processing_error')
+    .eq('id', eventId)
+    .maybeSingle();
+  if (procErrorResult.error) {
+    return {
+      success: false,
+      error: `processing_error lookup failed: ${maskPii(procErrorResult.error.message)}`,
+    };
+  }
+  const procError = procErrorResult.data?.processing_error as string | null | undefined;
+  if (procError) {
+    return {
+      success: false,
+      error: `Event processing failed: ${maskPii(procError)}`,
+    };
+  }
+
+  return { success: true, row: readBackResult.data as T };
+}
--- a/infrastructure/supabase/supabase/functions/manage-user/__tests__/deactivate-readback.test.ts
+++ b/infrastructure/supabase/supabase/functions/manage-user/__tests__/deactivate-readback.test.ts
@@ -1,0 +1,327 @@
+/**
+ * Unit tests for the Pattern A v2 read-back retrofit on `manage-user.deactivate`.
+ *
+ * Run with:
+ *   deno test --allow-net manage-user/__tests__/deactivate-readback.test.ts
+ *
+ * Scope: tests the shared `checkProjectionReadback` helper directly because
+ * the manage-user Edge Function delegates the Pattern A v2 contract to it.
+ * Verifying the helper's contract end-to-end (8 cases) gives us the same
+ * coverage as wiring up a full Edge Function request fixture, with less
+ * harness surface area.
+ *
+ * Coverage (8 cases, architect-revised PR-D plan review):
+ *   1. Happy path: emit + projection updated + no processing_error → success
+ *   2. Handler-throws: row missing + processing_error set → masked error
+ *   3. Race-safe: projection looks updated but processing_error set → masked error
+ *   4. Expected-state mismatch: row exists but doesn't match predicate → row treated as missing
+ *   5. Helper queries by captured event_id (NOT stream_id / latest event) → race-safe contract
+ *   6. Helper does NOT swallow `domain_events` query errors (RLS, missing GRANT)
+ *   7. Helper does NOT swallow projection-table query errors
+ *   8. PII masking on processing_error string (architect Q6)
+ *
+ * Auth-ban-failure and pre-emit-guard cases are covered by the Edge Function's
+ * pre-existing console.warn paths; they don't depend on the helper.
+ */
+
+import { assertEquals } from 'https://deno.land/std@0.220.1/assert/mod.ts';
+
+import { checkProjectionReadback } from '../../_shared/rpc-readback.ts';
+
+// =============================================================================
+// Mock Supabase client
+// =============================================================================
+
+interface MockQueryFixture {
+  /** Returns the data + error to surface for this query. */
+  data?: unknown;
+  error?: { message: string } | null;
+}
+
+interface MockClientConfig {
+  /** Map of (table → keyed fixture). */
+  tables: Record<string, MockQueryFixture>;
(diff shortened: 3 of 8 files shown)


infrastructure/supabase/supabase/functions/manage-user/index.ts
```diff
@@ -343,6 +351,55 @@ serve(async (req) => {
...
+          readback.error,
+        );
+        const errorResponse: ManageUserResponse = {
+          success: false,
+          error: readback.error,
+        };
+        return new Response(JSON.stringify(errorResponse), {
+          status: 500,
+          headers: { ...corsHeaders, 'Content-Type': 'application/json' },
+        });
```
lars-tice, line 404: **P1 — Response contract for read-back failure breaks parity with SQL Pattern A v2 + Edge Function standard error shape**

This returns HTTP **500** with a body shaped like `ManageUserResponse` (`{success: false, error}`). Two issues with that choice:

1. **Parity with SQL Pattern A v2.** Decision 2 in `adr-rpc-readback-pattern.md` makes envelope-handler-failures return **200 OK** with `{success: false, error}` for a reason: callers parse `data.success`, never an HTTP status. The SQL precedent at `api.delete_user` (and all 22 RPCs catalogued in the ADR) follow that convention. Going 500 here splits the pattern at the wire-tier port for no apparent gain — and per the helper's own docblock the wire-tier port is supposed to be *safer-with-same-shape*, not different-shape.

2. **Bypasses the Edge Function error chokepoint.** `_shared/error-response.ts` has `handleRpcError` / `createInternalError` precisely for event-processing failures: they emit `EdgeFunctionErrorResponse` (`{error, code: 'EVENT_PROCESSING_FAILED', correlation_id, details}`). The hand-rolled response here has no `code`, no `correlation_id` header, and a different body shape. Downstream `extractEdgeFunctionError` at `frontend/src/utils/edge-function-errors.ts:53` will surface `code: 'HTTP_ERROR'` (the fallback) instead of `'EVENT_PROCESSING_FAILED'`, and `correlationId: undefined` — losing the trace correlation on the exact error path operators need to follow up on.

**Recommended fix (pick one, prefer A):**

A. Return 200 + envelope to mirror SQL Pattern A v2:
```ts
return new Response(JSON.stringify({ success: false, error: readback.error }), {
  status: 200,
  headers: { ...corsHeaders, 'Content-Type': 'application/json' },
});
```

B. Keep 500 but use the standard chokepoint so `code` + `correlation_id` flow through:
```ts
return createErrorResponse(
  { error: readback.error, code: ErrorCodes.EVENT_PROCESSING_FAILED, status: 500, correlationId },
  corsHeaders,
);
```

(A) is the architecturally consistent answer; the helper's docblock already justifies it (`Event processing failed: <masked>` prefix; consumers parse the envelope).

infrastructure/supabase/supabase/functions/_shared/rpc-readback.ts
```diff
@@ -0,0 +1,158 @@
...
+ * The error string is structured for the frontend to surface verbatim in
+ * toast/banner UIs — the boundary helper does the masking so no caller is
+ * responsible for it.
+ */
+export async function checkProjectionReadback<T>(
+  client: AnyClient,
+  eventId: string,
+  table: string,
+  lookupKey: string,
+  lookupValue: string,
```
lars-tice: **P2 — `lookupValue: string` constrains reuse; sibling card and any non-UUID lookup will need a signature widening**

The parameter is typed `lookupValue: string`. UUID keys are strings, so deactivate works. But the docblock advertises this helper as the reuse target for the sibling reactivate card and the broader Edge Function port surface (PR description: "Designed for byte-equivalent reuse by the sibling `manage-user-reactivate-pattern-a-v2-retrofit`... `expectedState: { is_active: true }`").

Reuse is byte-equivalent in the immediate case, but the moment a future caller needs a non-string PK (e.g., composite/jsonb keys for projection tables with surrogate identity, or a non-text natural key), the signature has to widen and every existing call site has to be touched.

Low-cost forward-compat:
```ts
lookupValue: string | number,
```
or (more thorough):
```ts
lookupValue: unknown,
```
since `.eq()` accepts `unknown` at runtime anyway.

**Verdict**: P2, not blocking. UUID-only is acceptable as a documented current limitation if you'd rather keep the surface tight. If you keep `string`, please add a one-line note in the docblock: "Helper is UUID-string-keyed; signature widening will be needed if a non-UUID-keyed projection adopts Pattern A v2."

infrastructure/supabase/supabase/functions/manage-user/__tests__/deactivate-readback.test.ts
```diff
@@ -0,0 +1,327 @@
...
+  );
+
+  assertEquals(result.success, false);
+  if (!result.success) {
+    assertEquals(
+      result.error,
+      'Event processing failed: projection read-back returned no row',
+    );
+  }
+});
```
lars-tice, line 191: **P3 — Test 4 does not actually exercise the predicate-mismatch branch; it tests NOT FOUND with `processing_error: null`**

The mock `chain.eq()` ignores its arguments and `chain.maybeSingle()` returns the fixture's `data` verbatim. So fixturing `users: { data: null }` does not simulate the case "row exists in DB but `.eq('is_active', false)` predicate excludes it" — it just simulates "row absent." The test then asserts the fall-through message `'projection read-back returned no row'`, which is correct for the NOT FOUND code path.

In other words, case 4 is mechanically indistinguishable from case 2 with `processing_error: null` — it doesn't add coverage for predicate-vs-row-existence distinction. The real predicate-mismatch behavior (`.eq('is_active', false)` filtering a row that is still `is_active: true`) is exercised at integration time, not in this unit suite.

**Action**: either upgrade the mock so `.eq()` is filter-aware (chain remembers `.eq(col, val)` calls, `maybeSingle()` checks fixture against them), or relabel the test from "Expected-state mismatch" to "NOT FOUND with no processing_error → fallback message" so future readers don't over-trust it. Coverage on the NOT FOUND fallback message is real either way; only the test name is misleading.

P3 — docblock/rename suffices.

frontend/src/types/client.types.ts
```diff
@@ -706,6 +706,14 @@ export type ClientProjectionRow = Omit<
...
   success: boolean;
   error?: string;
+  /**
+   * Captured event_id from the underlying `emit_domain_event` call. Additive
+   * on success responses for audit-log deep-linking. Optional — consumers
+   * are not required to surface it. First introduced 2026-05-12 alongside
+   * the `manage-user.deactivate` Pattern A v2 read-back retrofit; available
+   * uniformly on every envelope-returning service call going forward.
+   */
+  eventId?: string;
```
lars-tice, line 717: **P3 — `ClientRpcEnvelope.eventId` has no current emitter; reads as "reserved field"**

Uniformity with `UserRpcEnvelope` is the stated rationale, and that's defensible. But no `api.*` client RPC currently populates `eventId` on the success envelope (verified via grep: `register_client`, `update_client`, `add_client_*`, `update_client_*`, `change_client_placement`, `assign_client_contact`, etc. — none return `eventId` in their `jsonb_build_object` success body, and no client service consumer reads `env.eventId`).

For a reader, the docblock saying "Captured event_id from the underlying `emit_domain_event` call" reads like an active contract. It isn't — it's a reservation.

**Suggestion**: tighten the docblock to make the reservation explicit so a future engineer doesn't (a) assume client RPCs already emit it and stop verifying, or (b) decide it's dead and remove it:

```ts
/**
 * Captured event_id from the underlying `emit_domain_event` call. **Reserved
 * for future use** — no `api.*` client RPC currently emits this field. Added
 * 2026-05-12 alongside `UserRpcEnvelope.eventId` for cross-service
 * uniformity; will be populated as individual client write RPCs adopt the
 * audit-deep-linking pattern.
 */
eventId?: string;
```

P3 — purely doc, in-scope inline fix.

infrastructure/supabase/supabase/functions/manage-user/__tests__/deactivate-readback.test.ts
```diff
@@ -0,0 +1,327 @@
...
+
+  const wrongQueries = eqCalls.filter(
+    (call) => call.column === 'stream_id' || call.column === 'stream_type',
+  );
+  assertEquals(
+    wrongQueries.length,
+    0,
+    'helper must NOT query domain_events by stream_id (race-unsafe)',
+  );
+});
```
lars-tice, line 235: **P3 — Test 5's race-safety assertion checks the wrong table coupling**

The filter `eqCalls.filter(call => call.column === 'id' && call.value === EVENT_ID)` correctly catches the event-id query, but the helper also calls `.eq('id', USER_ID)` on the `users` table — both share the column name `id`. The mock's `onEq` callback is table-agnostic. The test therefore cannot distinguish "helper queried `domain_events` by event_id" from "helper queried `users` table by an id that happened to equal `EVENT_ID`."

The far-fetched bug class this misses: "helper queries `users` table by event_id (using event_id as lookupValue) and `domain_events` by user_id." That's not a realistic regression — the code paths don't share that variable confusion — so this is genuinely P3.

**Minor improvement**: extend the mock to thread the table name through:

```ts
onEq?: (table: string, column: string, value: unknown) => void;
// in from():
return {
  select(...) { return chain; },
  eq(col, val) { onEq(table, col, val); return chain; },
  ...
};
```

Then assert `(table === 'domain_events' && column === 'id' && value === EVENT_ID)`. The `wrongQueries` check on `stream_id`/`stream_type` already provides 80% of the race-safety guarantee — this just closes the last 20%.

P3 — non-blocking.

lars-tice (commented):
## Verdict: APPROVE WITH IN-PR FIXES — 1 P1, 1 P2, 3 P3

First Edge Function adopter of Pattern A v2 is well-shaped and architecturally aligned with the established SQL precedent (`api.delete_user` two-check form). The new shared helper `_shared/rpc-readback.ts` is faithful to the SQL pattern, correctly captures the race-safe event_id semantics, and applies `maskPii` at the right boundary. Reactivate is intentionally deferred; the scope discipline mirrors the user's "address inline, not via cards" rule applied correctly (sibling card is genuinely separate work, not deferred cleanup).

The PR is **mostly correct**, but the wire-tier response shape for the read-back failure path is inconsistent with both Pattern A v2 SQL parity and the Edge Function error chokepoint — fixable in a few lines.

### Action list (must be addressed before merge)

**P1 — Response shape for read-back failure (inline F1, `manage-user/index.ts:390-397`)**
- Pick one of two options:
  - (A) **Preferred**: return HTTP 200 + `{success: false, error: readback.error}` to match SQL Pattern A v2 parity.
  - (B) Keep 500 but route through `_shared/error-response.ts` (`createErrorResponse({code: ErrorCodes.EVENT_PROCESSING_FAILED, ...})`) so the frontend receives `code: 'EVENT_PROCESSING_FAILED'` and `correlation_id` instead of the current `code: 'HTTP_ERROR'` (fallback) and `correlationId: undefined`.
- Either way the SQL Pattern A v2 invariant "callers parse `data.success`, never an HTTP status" should be preserved. Today's 500 + non-standard body splits the pattern at the wire tier for no apparent gain.

**P2 — `lookupValue: string` constrains forward reuse (inline F2, `_shared/rpc-readback.ts:97`)**
- Either widen to `string | number` / `unknown`, or add a docblock note explicitly limiting the helper to UUID-string-keyed projections. Today the signature is tighter than the helper's stated reuse mandate.

### Notes (P3 — non-blocking, fix opportunistically)

- **P3 — Test 4 fidelity** (inline F3, `deactivate-readback.test.ts:157-184`): the predicate-mismatch case mechanically tests the NOT FOUND fallback message, not actual predicate filtering. Either upgrade the mock to be `.eq()`-aware or rename the test so future readers don't over-trust its scope.
- **P3 — Test 5 table coupling** (inline F5, `deactivate-readback.test.ts:186-224`): `onEq` is table-agnostic, so the race-safety assertion can't distinguish `users.id = EVENT_ID` from `domain_events.id = EVENT_ID`. Thread the table name through `onEq` to close the gap.
- **P3 — `ClientRpcEnvelope.eventId` reservation** (inline F4, `client.types.ts:706-717`): no client RPC currently populates this. Cross-service uniformity is fine as a rationale; tighten the docblock to say "Reserved for future use — no `api.*` client RPC currently emits this" so it doesn't read like an active contract that's been overlooked.

### Architecture Review Checklist

- **CQRS Standards** — PASS. No projection-direct queries introduced; helper reads `domain_events.processing_error` by primary key, which is the correct race-safe lookup. Frontend service still calls Edge Function via `functions.invoke`, no `.schema('api').rpc(...)` direct calls.
- **Naming Conventions** — PASS. Helper name (`checkProjectionReadback`), file location (`_shared/rpc-readback.ts`), `DEPLOY_VERSION` tag, and ManageUserResponse field name (`eventId`) all align with established conventions.
- **Design Patterns** — PASS. Two-check Pattern A v2 form is the canonical pattern; helper extraction matches the SQL precedent's structure. Wire-tier "safer than SQL intra-transaction" claim in the docblock is correct (post-emit RPC commit means `processing_error` is durably visible to the next read).
- **data-testid Attributes** — N/A. No UI changes.
- **AsyncAPI Event Registration** — N/A. No new events; emits pre-existing `user.deactivated`.
- **Type Generation — No Anonymous Types** — PASS. `ProjectionReadbackResult<T>` is named/exported; `ManageUserResponse` named; envelope changes use named `UserRpcEnvelope` / `ClientRpcEnvelope` types; `expectedState: Record<string, unknown>` is the appropriate generic-bag form for the predicate.
- **Observability, Tracing & Monitoring** — NEEDS ATTENTION (rolls into P1). On the read-back failure path, the current 500 + non-standard body bypasses the `correlation_id` header path that operators use to trace the failure. The P1 fix (route through `createErrorResponse` OR mirror SQL Pattern A's 200 + envelope) closes this.
- **Error Surfacing to UI** — PASS in the happy path; **NEEDS ATTENTION on read-back failure**. The 500 body's missing `code` field means `errorDetails.code` falls back to `'HTTP_ERROR'` in the frontend instead of `'EVENT_PROCESSING_FAILED'`. Toast/banner can still render the message, but consumers can't programmatically branch on the code. Closes with the same P1 fix.

### Patterns explicitly verified

- Rule 7.1 (handler reference files): N/A — no handler/router touched.
- Rule 15 (baseline overload audit): N/A — no new `api.*` function.
- Rule 16 (no identifier interpolation): handler unchanged; helper masks `processing_error` defensively at the boundary regardless.
- Rule 17 (RPC shape registry): N/A — no `api.*` RPC added.
- DROP+CREATE re-tag rule: N/A — no migration in this PR.
- `apiRpc` vs `apiRpcEnvelope`: Edge Function path uses `functions.invoke`, correct. Frontend `deactivateUser` passes `eventId` through additively, consistent with the M3 envelope contract.

### Considered but did not post (judgment calls — surfacing for visibility)

- **Test 8 (PII masking)** is genuinely strong (asserts UUID, email, and structural-PG-detail patterns all masked). No finding.
- **Auth-ban failure as soft warning** (lines 421-445) — Pre-existing, out of scope per the PR description. The new read-back is ordered correctly (BEFORE the ban call) so a read-back failure won't leave a banned-but-not-deactivated state. No finding; the existing scope discipline is correct.
- **Order of read-back and `auth.admin.updateUserById`** — Considered flagging that the auth ban happens after the read-back succeeds. This is correct (don't pretend to deactivate if the projection didn't update), so no finding.
- **`AnyClient = SupabaseClient<any, any, any>`** — Tradeoff matches the established `_shared/types.ts:AnySchemaSupabaseClient` pattern. No finding.
- **`reactivate` not surfacing `eventId`** (`SupabaseUserCommandService.ts:421`) — intentional per the sibling card's scope. The pattern is "additive when the underlying caller returns it," which deactivate now does and reactivate will in the next PR. No finding.
- **Edge Function `code` field could be added to the read-back failure body without changing status** — folded into the P1 action list as option B rather than a separate finding.

infrastructure/supabase/supabase/functions/manage-user/index.ts
lars-tice: Addressed in commit a4efd0f4 — Option A (SQL Pattern A v2 parity). Response is now `HTTP 200 + {success: false, error, eventId}` on read-back failure. Callers parse `data.success` regardless of wire tier; same frontend envelope-unwrap logic works for both SQL RPC and Edge Function paths. `eventId` is also surfaced on the failure response for audit-log deep-linking even on failed operations.

infrastructure/supabase/supabase/functions/_shared/rpc-readback.ts
lars-tice: Addressed in commit a4efd0f4 — widened to `lookupValue: string | number` with updated docblock. `.eq()` accepts both at runtime; the helper now generalizes to surrogate-integer-keyed projections without forcing callers to coerce. If a future caller needs composite/jsonb keys we can widen further or add an overload, but `string | number` is the conservative forward-compat that costs nothing for the current UUID-keyed adopters.

infrastructure/supabase/supabase/functions/manage-user/__tests__/deactivate-readback.test.ts
lars-tice: Addressed in commit a4efd0f4 — chose the rename path (lower scope than upgrading the mock to be `.eq()`-aware). Test 4 is now titled "NOT FOUND with no processing_error → fallback message" with an explanatory comment noting that true predicate-mismatch behavior is verified at integration time. Coverage on the fallback message stays real; the scope is no longer misleading.

frontend/src/types/client.types.ts
lars-tice: Addressed in commit a4efd0f4 — docblock now says "**Reserved for future use** — no `api.*` client RPC currently emits this field on its success envelope. Added 2026-05-12 alongside `UserRpcEnvelope.eventId` for cross-service uniformity; will be populated as individual client write RPCs adopt the audit-deep-linking pattern." Prevents both failure modes flagged (assume-already-emitted vs assume-dead).

infrastructure/supabase/supabase/functions/manage-user/__tests__/deactivate-readback.test.ts
lars-tice: Addressed in commit a4efd0f4 — threaded `table` through `onEq(table, column, value)`. Test 5 now asserts `call.table === 'domain_events' && call.column === 'id' && call.value === EVENT_ID` specifically, closing the table-coupling gap. The negative-case filter on `stream_id`/`stream_type` is now also table-scoped to `domain_events`. Both halves of the race-safety guarantee covered.

lars-tice (commented):
## Follow-up verdict: ALL ADDRESSED — READY TO MERGE

Verified commit `a4efd0f4` against all 5 prior findings. No regressions found, no new architectural concerns introduced. Deno tests (8/8), TypeScript, and `lint --max-warnings 0` all green.

### Finding-by-finding verification

| Finding | Status | Evidence |
|---|---|---|
| **P1** — Read-back failure missing `code`/`correlation_id`; inconsistent with SQL Pattern A v2 | ✅ ADDRESSED | `infrastructure/supabase/supabase/functions/manage-user/index.ts:386–404` — Chose **Option A** (architect's preferred). Returns HTTP 200 + `{success: false, error, eventId}`. Aligns with `adr-rpc-readback-pattern.md` Decision 5 ("Under Pattern A, PostgREST always returns 200 OK with `{success, error}`, so there is no `code` field to read"). With HTTP 200 the SDK no longer raises `FunctionsHttpError`; the frontend `extractEdgeFunctionError` fallback path (`code: 'HTTP_ERROR'`) is bypassed entirely — flow now goes through `SupabaseUserCommandService.ts:340` (`if (!data?.success)`). The `eventId` surfaced on failure is a free win for audit-log deep-linking. Note: helper-level tests (8/8 pass) verify the `checkProjectionReadback` `{success, error}` shape that feeds the wire response, but no test asserts the literal HTTP 200 + body shape end-to-end. Acceptable because the wire-tier transformation is two lines of glue code (`status: 200`, JSON.stringify the typed `ManageUserResponse`) — adding a fetch-mock test would be tarpit work for negligible additional coverage. |
| **P2** — `lookupValue: string` constrains forward reuse | ✅ ADDRESSED | `infrastructure/supabase/supabase/functions/_shared/rpc-readback.ts:78–81, 100` — Widened to `string \| number` with docblock note explaining the two key spaces (UUID-keyed vs surrogate-integer-keyed projections). Backward-compatible for the sibling reactivate retrofit (`expectedState: { is_active: true }`, `lookupValue` still a UUID string). |
| **P3 F3** — Test 4 mechanically tests NOT FOUND fallback, not predicate filtering | ✅ ADDRESSED | `infrastructure/supabase/supabase/functions/manage-user/__tests__/deactivate-readback.test.ts:162–191` — Renamed to "`NOT FOUND with no processing_error → fallback message`". Inline comment explicitly states the scope ("the mock's `.eq()` is filter-blind, so this test mechanically exercises the NOT FOUND branch") and defers true predicate-mismatch coverage to integration time. Chose the rename path (lower effort) over the alternative (make the mock `.eq()`-aware) — defensible for a unit test layer. |
| **P3 F4** — `ClientRpcEnvelope.eventId` docblock should say "Reserved for future use" | ✅ ADDRESSED | `frontend/src/types/client.types.ts:709–717` — Docblock now leads with "**Reserved for future use** — no `api.*` client RPC currently emits this field on its success envelope." Reads correctly as a forward-compat reservation, not an active contract. |
| **P3 F5** — Test 5 `onEq` callback table-agnostic; can't distinguish `users.id` from `domain_events.id` | ✅ ADDRESSED | `deactivate-readback.test.ts:50, 63, 193–235` — `onEq` signature widened to `(table, column, value)`; spy threading propagates the table name; positive assertion at L214–217 now filters on `table === 'domain_events' && column === 'id' && value === EVENT_ID`; negative belt at L224–229 also filters on `table === 'domain_events'` for `stream_id`/`stream_type`. Closes the architect-flagged gap completely. |

### Regression scan on the fix commit

- No new dependencies added.
- Public type widening on `checkProjectionReadback.lookupValue` (`string → string | number`) — covariant for call sites, backward-compatible.
- `ManageUserResponse.eventId` is `?: string` and additive — no caller break.
- Frontend `client.types.ts` change is a comment-only edit; no API surface change.
- Deno 8/8 ok, frontend typecheck/lint green.

### Architectural note (no action — observation only)

The author's choice of Option A is the architecturally consistent one: SQL Pattern A v2 RPCs and the Edge Function port now share the same wire contract (`HTTP 200 + {success, error?, ...}`). Frontend callers can route both shapes through `apiRpcEnvelope<T>`-equivalent semantics without per-tier branching. The `eventId` addition on failure responses is a small surface improvement that will pay off once the audit-log deep-link UI lands.

`manage-user.deactivate` is the first wire-tier adopter of Pattern A v2; the helper (`checkProjectionReadback`) is byte-ready for the sibling reactivate retrofit (`expectedState: { is_active: true }`, same `string | number` `lookupValue`). Recommended that the next sibling card lift the deactivate test file as a template.

### Verdict

**ALL 5 ADDRESSED — READY TO MERGE.** No remaining architect-side blockers.

PR #61, opened 2026-05-12
fix(edge-functions): close schema-mismatch bug class — manage-user + accept-invitation

lars-tice:
## Summary

**Hotfix expanded** per architect plan-review on the 2026-05-12 deactivate UAT failure. The original hotfix closed the loud failure in `_shared/rpc-readback.ts`; this expansion closes a SECOND **silent-failure** instance in `accept-invitation` + addresses the architect's 8 plan-review findings inline (no follow-up cards).

**Bug class**: Edge Function clients constructed with `db: { schema: 'api' }` for RPC ergonomics + naked `.from(<public-table>)` reads → PostgREST resolves the query as `api.<table>` (missing) → silent or loud failure depending on whether the caller swallows the error.

**Architect verdict**: APPROVE WITH IN-PR FIXES — all 8 findings (1 P1, 3 P2, 4 P3) addressed inline per "no deferral to cards" rule.

## F1 (P1 BLOCKER) — accept-invitation silent re-emission

`accept-invitation/index.ts:515-531` had the same bug class with **WORSE** semantics than manage-user:
- L515-519: `supabase.from('users').select('deleted_at')` — resolves as `api.users` → silent failure → `userRow = null` → `isDeleted = false`
- L527-531: `supabase.from('user_roles_projection').select('id')` — resolves as `api.user_roles_projection` → silent failure → `existingRoles = null`
- L536-539: `console.warn(...) // Continue` — swallows the error
- L541-544: `isExistingUser = false` (no rows because queries failed) → emits `user.created` for EVERY existing-user OAuth/SSO invitation accept

Impact: every existing-user SSO/OAuth invitation accept polluted the audit trail with a duplicate `user.created` event. Handler re-runs; correlation_id may collide.

**Fix**: extracted to `checkExistingUserPath(client, userId)` named export with explicit `.schema('public')` on both queries. New 6-case Deno regression test in `accept-invitation/__tests__/existing-user-check-schema.test.ts` uses the same `.from()`-spy invariant pattern as `manage-user` Test 9.

## F2 — Test 9 upgraded to `.from()`-spy invariant

Replaced count-based `schemaCalls.length >= 2` with the architectural invariant: every `.from(table)` MUST have a preceding `.schema('public')`. Catches future regressions regardless of query count.

## F3 — Schema-pinning invariant promoted to JSDoc

`_shared/rpc-readback.ts` now has a "# Schema-pinning invariant" section in the top-of-file docblock explaining the rule + linking to the regression test pattern. Visible to IDE hover and to future Pattern A v2 adopters (reactivate retrofit lifts this unchanged).

## F4 — Type-alias dedup

Dropped local `AnyClient` in `rpc-readback.ts`; imports `AnySchemaSupabaseClient` from `_shared/types.ts`.

## F5 — Audited remaining Edge Functions (all clean)

Verified `invite-user`, `organization-bootstrap`, `organization-delete`, `workflow-status` — none have `.from()` calls (pure-RPC). **Bug class is contained** to `_shared/rpc-readback.ts` + `accept-invitation/index.ts`, both now fixed.

## F6 + F8 — Pre-deploy ritual gap + tasks.md template

8/8 unit tests + lint + typecheck + build + successful deploy all green; first signal of failure was UAT Test 1. Class of bugs missed: config-dependent SDK behavior (schema, RLS, GRANT). Added "Deploy to dev + smoke against real Supabase + paste log evidence" pre-PR-open task line to the active card's `tasks.md` as the template demonstration. Self-acknowledged: THIS card's hotfix cycle didn't follow that discipline.

## F7 — Rule-pinning deferred to 3rd precedent

The `supabaseAdmin` construction asymmetry (schema='api' but used for cross-schema reads) — helper-side `.schema('public')` pinning is the right architectural answer. **Defer** rule-pinning to `.claude/skills/infrastructure-guidelines/SKILL.md` / `infrastructure/supabase/CLAUDE.md` until a 3rd Edge Function shows the same pattern. Speculative rule-writing at 2 precedents would overfit. **Deferral memorialized in MEMORY.md** so the decision isn't lost.

## Test plan

- [x] `deno test --allow-net manage-user/__tests__/ accept-invitation/__tests__/` — **26/26 pass** (9 deactivate + 6 existing-user + 11 phone-id-resolution)
- [x] `npm run typecheck` — green
- [x] `npm run lint -- --max-warnings 0` — green
- [x] No frontend changes needed (wire contract unchanged)
- [ ] Manual smoke (post-deploy): redo UAT Test 1 for deactivate; trigger an existing-user OAuth/SSO invitation accept and verify `user.created` is NOT re-emitted (`SELECT count(*) FROM domain_events WHERE stream_id = '<user_id>' AND event_type = 'user.created'` returns 1, not 2).

## Memorialization (MEMORY.md, in this PR)

- New "Stale Content Pitfalls" entry — Edge Function schema-pinning rule + regression test pattern
- New "Stale Content Pitfalls" entry — Pre-deploy ritual gap for config-dependent SDK behavior
- Rule-pinning deferral preserved with explicit trigger ("3rd precedent")

### DIFF
--- a/infrastructure/supabase/supabase/functions/_shared/rpc-readback.ts
+++ b/infrastructure/supabase/supabase/functions/_shared/rpc-readback.ts
@@ -104,7 +104,18 @@ export async function checkProjectionReadback<T>(
   // If NOT FOUND: query `processing_error` on the captured event_id and
   // surface the masked failure reason. Surface query errors verbatim
   // (RLS denials, missing GRANTs would be silent-failure modes if swallowed).
-  let readBackQuery = client.from(table).select('*').eq(lookupKey, lookupValue);
+  //
+  // Schema note: Edge Function service-role clients (e.g. `supabaseAdmin` in
+  // `manage-user/index.ts`) are typically constructed with `db: { schema: 'api' }`
+  // for RPC ergonomics. Projection tables and `domain_events` live in `public`,
+  // so the read-back must explicitly target the public schema via
+  // `.schema('public')`. Returned bug from first deployment 2026-05-12:
+  // "Could not find the table 'api.users' in the schema cache".
+  let readBackQuery = client
+    .schema('public')
+    .from(table)
+    .select('*')
+    .eq(lookupKey, lookupValue);
   for (const [column, value] of Object.entries(expectedState)) {
     readBackQuery = readBackQuery.eq(column, value);
   }
@@ -118,6 +129,7 @@ export async function checkProjectionReadback<T>(
 
   if (!readBackResult.data) {
     const procErrorResult = await client
+      .schema('public')
       .from('domain_events')
       .select('processing_error')
       .eq('id', eventId)
@@ -139,6 +151,7 @@ export async function checkProjectionReadback<T>(
   // when the read-back looks fine. Catches the partial-update-then-throw
   // window where the handler successfully wrote some state before raising.
   const procErrorResult = await client
+    .schema('public')
     .from('domain_events')
     .select('processing_error')
     .eq('id', eventId)
--- a/infrastructure/supabase/supabase/functions/_shared/rpc-readback.ts
+++ b/infrastructure/supabase/supabase/functions/_shared/rpc-readback.ts
@@ -13,6 +13,23 @@
  *    handler-update-then-throw window AND the case where the projection
  *    looked updated for an unrelated reason.
  *
+ * # Schema-pinning invariant
+ *
+ * Edge Function service-role clients (e.g. `supabaseAdmin` in `manage-user/index.ts`)
+ * are typically constructed with `db: { schema: 'api' }` for RPC ergonomics.
+ * Projection tables and `domain_events` live in **`public`**, so every `.from()`
+ * call inside this helper MUST be preceded by `.schema('public')`. Pre-fix
+ * (2026-05-12 hotfix, PR #61), the helper called `client.from(...)` directly
+ * and PostgREST resolved every query as `api.<table>`:
+ *
+ *     Could not find the table 'api.users' in the schema cache
+ *
+ * The helper's regression test (Test 9 in `manage-user/__tests__/`) enforces
+ * the `.from()`-spy invariant: every `.from()` MUST have a preceding
+ * `.schema('public')` call. The architect PR #58 + PR #61 review codified
+ * this as the chokepoint pattern for any future Pattern A v2 Edge Function
+ * adopter (e.g. the reactivate retrofit, which lifts this helper unchanged).
+ *
  * # Transactional model — SQL RPC vs Edge Function
  *
  * In the SQL precedent (`api.delete_user`, `api.update_role`, etc.) the
@@ -45,17 +62,8 @@
  * - First adopter: `manage-user.deactivate` (this PR)
  */
 
-import type { SupabaseClient } from 'https://esm.sh/@supabase/supabase-js@2';
 import { maskPii } from './maskPii.ts';
-
-/**
- * Loose type alias matching the AnySchemaSupabaseClient pattern used elsewhere
- * in `_shared` and the frontend SDK boundary. Required because `supabaseAdmin`
- * is configured with the default 'public' schema but reads from `api` and
- * other tables.
- */
-// deno-lint-ignore no-explicit-any
-type AnyClient = SupabaseClient<any, any, any>;
+import type { AnySchemaSupabaseClient } from './types.ts';
 
 export type ProjectionReadbackResult<T> =
   | { success: true; row: T }
@@ -93,7 +101,7 @@ export type ProjectionReadbackResult<T> =
  * responsible for it.
  */
 export async function checkProjectionReadback<T>(
-  client: AnyClient,
+  client: AnySchemaSupabaseClient,
   eventId: string,
   table: string,
   lookupKey: string,
--- a/infrastructure/supabase/supabase/functions/manage-user/__tests__/deactivate-readback.test.ts
+++ b/infrastructure/supabase/supabase/functions/manage-user/__tests__/deactivate-readback.test.ts
@@ -48,27 +48,43 @@ interface MockClientConfig {
    * `domain_events.id = EVENT_ID` — both share the column name `id`.
    */
   onEq?: (table: string, column: string, value: unknown) => void;
+  /**
+   * Callback invoked on every `.schema(name)` call. Regression guard for the
+   * 2026-05-12 hotfix where the helper called `client.from('users')` directly
+   * — `users` lives in `public`, but the Edge Function's `supabaseAdmin` is
+   * configured with `schema: 'api'` for RPC ergonomics, so PostgREST resolved
+   * the query as `api.users` (missing). Fix: explicit `.schema('public')` on
+   * every projection / domain_events query.
+   */
+  onSchema?: (schema: string) => void;
 }
 
 function makeMockClient(config: MockClientConfig) {
   const onEq = config.onEq ?? (() => {});
+  const onSchema = config.onSchema ?? (() => {});
+  const fromImpl = (table: string) => {
+    const fixture = config.tables[table] ?? { data: null, error: null };
+    const chain = {
+      select(_columns: string) {
+        return chain;
+      },
+      eq(column: string, value: unknown) {
+        onEq(table, column, value);
+        return chain;
+      },
+      async maybeSingle() {
+        return { data: fixture.data, error: fixture.error ?? null };
+      },
+    };
+    return chain;
+  };
+  const schemaImpl = (schema: string) => {
+    onSchema(schema);
+    return { from: fromImpl };
+  };
   return {
-    from(table: string) {
-      const fixture = config.tables[table] ?? { data: null, error: null };
-      const chain = {
-        select(_columns: string) {
-          return chain;
-        },
-        eq(column: string, value: unknown) {
-          onEq(table, column, value);
-          return chain;
-        },
-        async maybeSingle() {
-          return { data: fixture.data, error: fixture.error ?? null };
-        },
-      };
-      return chain;
-    },
+    schema: schemaImpl,
+    from: fromImpl,
     // deno-lint-ignore no-explicit-any
   } as any;
 }
@@ -289,6 +305,43 @@ Deno.test('7. Helper does NOT swallow projection-table query errors', async () =
       `expected read-back query failure, got: ${result.error}`,
     );
   }
+});
+
+Deno.test('9. Schema regression guard: queries explicitly target public schema (2026-05-12 hotfix)', async () => {
+  // The Edge Function's service-role client is constructed with `db: { schema: 'api' }`
+  // for RPC ergonomics. Before the hotfix the helper called `client.from(table)`
+  // directly, which resolved as `api.<table>` — but `users` and `domain_events`
+  // live in `public`, so PostgREST returned "Could not find the table 'api.users'
+  // in the schema cache". The fix is explicit `.schema('public')` on every
+  // projection / domain_events query. This test fails if a future refactor
+  // drops the `.schema('public')` calls.
+  const schemaCalls: string[] = [];
+  const client = makeMockClient({
+    tables: {
+      users: { data: { id: USER_ID, is_active: false } },
+      domain_events: { data: { processing_error: null } },
+    },
+    onSchema: (schema) => {
+      schemaCalls.push(schema);
+    },
+  });
+
+  await checkProjectionReadback(client, EVENT_ID, 'users', 'id', USER_ID, {
+    is_active: false,
+  });
+
+  // Happy path issues exactly two queries (projection read-back + race-safe
+  // processing_error check). Both must explicitly route through public schema.
+  assertEquals(
+    schemaCalls.every((s) => s === 'public'),
+    true,
+    `every .schema() call must target 'public'; got: ${JSON.stringify(schemaCalls)}`,
+  );
+  assertEquals(
+    schemaCalls.length >= 2,
+    true,
+    'helper must call .schema() at least twice (read-back + processing_error check)',
+  );
 });
 
 Deno.test('8. PII masking on processing_error string (architect Q6)', async () => {
--- a/infrastructure/supabase/supabase/functions/manage-user/__tests__/deactivate-readback.test.ts
+++ b/infrastructure/supabase/supabase/functions/manage-user/__tests__/deactivate-readback.test.ts
@@ -49,20 +49,32 @@ interface MockClientConfig {
    */
   onEq?: (table: string, column: string, value: unknown) => void;
   /**
-   * Callback invoked on every `.schema(name)` call. Regression guard for the
-   * 2026-05-12 hotfix where the helper called `client.from('users')` directly
-   * — `users` lives in `public`, but the Edge Function's `supabaseAdmin` is
-   * configured with `schema: 'api'` for RPC ergonomics, so PostgREST resolved
-   * the query as `api.users` (missing). Fix: explicit `.schema('public')` on
-   * every projection / domain_events query.
+   * Callback invoked on every `.schema(name)` call. Used in conjunction with
+   * `onFrom` to enforce the schema-pinning invariant: every `.from()` call
+   * MUST be preceded by `.schema('public')`. Regression guard for the
+   * 2026-05-12 hotfix.
    */
   onSchema?: (schema: string) => void;
+  /**
+   * Callback invoked on every `.from(table)` call with the table name and the
+   * most-recent `.schema()` value (or `null` if no `.schema()` preceded it).
+   * Architectural invariant: in this helper, `currentSchema` MUST be
+   * `'public'` at every `.from()` call site — `null` or any other value
+   * means a schema-pinning regression. The architect's PR #61 review F2
+   * explicitly preferred this `.from()`-spy invariant over count-based
+   * assertions because it catches ANY future violation regardless of how
+   * the helper's query count changes.
+   */
+  onFrom?: (table: string, currentSchema: string | null) => void;
 }
 
 function makeMockClient(config: MockClientConfig) {
   const onEq = config.onEq ?? (() => {});
   const onSchema = config.onSchema ?? (() => {});
-  const fromImpl = (table: string) => {
+  const onFrom = config.onFrom ?? (() => {});
+
+  const fromImpl = (table: string, currentSchema: string | null) => {
+    onFrom(table, currentSchema);
     const fixture = config.tables[table] ?? { data: null, error: null };
     const chain = {
       select(_columns: string) {
@@ -78,13 +90,19 @@ function makeMockClient(config: MockClientConfig) {
     };
     return chain;
   };
+
   const schemaImpl = (schema: string) => {
     onSchema(schema);
-    return { from: fromImpl };
+    // Returning a fresh `{from}` scope: after this `.schema(...)`, the next
+    // `.from(table)` will see `schema` in `currentSchema`. If a caller skips
+    // `.schema()` and goes straight to `client.from(...)`, currentSchema is
+    // null — which the schema-pinning test invariant treats as a regression.
+    return { from: (table: string) => fromImpl(table, schema) };
   };
+
   return {
     schema: schemaImpl,
-    from: fromImpl,
+    from: (table: string) => fromImpl(table, null),
     // deno-lint-ignore no-explicit-any
   } as any;
 }
@@ -307,22 +325,27 @@ Deno.test('7. Helper does NOT swallow projection-table query errors', async () =
   }
 });
 
-Deno.test('9. Schema regression guard: queries explicitly target public schema (2026-05-12 hotfix)', async () => {
-  // The Edge Function's service-role client is constructed with `db: { schema: 'api' }`
-  // for RPC ergonomics. Before the hotfix the helper called `client.from(table)`
-  // directly, which resolved as `api.<table>` — but `users` and `domain_events`
-  // live in `public`, so PostgREST returned "Could not find the table 'api.users'
-  // in the schema cache". The fix is explicit `.schema('public')` on every
-  // projection / domain_events query. This test fails if a future refactor
-  // drops the `.schema('public')` calls.
-  const schemaCalls: string[] = [];
+Deno.test('9. Schema-pinning invariant: every .from() preceded by .schema("public") (PR #61 F2)', async () => {
+  // Architect PR #61 F2: upgraded from a count-based `>= 2` check to the
+  // `.from()`-spy invariant. Architectural invariant: ANY `.from()` call in
+  // this helper without a preceding `.schema('public')` is a regression,
+  // regardless of how many queries the helper makes. Catches any future
+  // refactor that drops the `.schema('public')` chain on a new code path.
+  //
+  // The Edge Function's service-role client is constructed with
+  // `db: { schema: 'api' }` for RPC ergonomics. Before the hotfix the helper
+  // called `client.from(table)` directly, which resolved as `api.<table>` —
+  // but `users` and `domain_events` live in `public`, so PostgREST returned
+  // "Could not find the table 'api.users' in the schema cache".
+  const fromCallLog: Array<{ table: string; schema: string | null }> = [];
+
   const client = makeMockClient({
     tables: {
       users: { data: { id: USER_ID, is_active: false } },
       domain_events: { data: { processing_error: null } },
     },
-    onSchema: (schema) => {
-      schemaCalls.push(schema);
+    onFrom: (table, currentSchema) => {
+      fromCallLog.push({ table, schema: currentSchema });
     },
   });
 
@@ -330,17 +353,23 @@ Deno.test('9. Schema regression guard: queries explicitly target public schema (
     is_active: false,
   });
 
-  // Happy path issues exactly two queries (projection read-back + race-safe
-  // processing_error check). Both must explicitly route through public schema.
+  // Invariant: every .from() call MUST have been preceded by .schema('public').
+  // null = bare client.from(), any other value = wrong schema. Both fail.
+  for (const call of fromCallLog) {
+    assertEquals(
+      call.schema,
+      'public',
+      `.from('${call.table}') called without preceding .schema('public') — schema-pinning regression`,
+    );
+  }
+
+  // Belt: at least 2 queries (projection read-back + race-safe processing_error).
+  // Architectural floor; helpers may add more chains in the future and that's fine
+  // as long as each one continues to be schema-pinned (asserted above).
   assertEquals(
-    schemaCalls.every((s) => s === 'public'),
+    fromCallLog.length >= 2,
     true,
-    `every .schema() call must target 'public'; got: ${JSON.stringify(schemaCalls)}`,
-  );
-  assertEquals(
-    schemaCalls.length >= 2,
-    true,
-    'helper must call .schema() at least twice (read-back + processing_error check)',
+    `helper must call .from() at least twice; got ${fromCallLog.length}`,
   );
 });
 
--- a/infrastructure/supabase/supabase/functions/manage-user/index.ts
+++ b/infrastructure/supabase/supabase/functions/manage-user/index.ts
@@ -42,7 +42,7 @@ import { buildEventMetadata } from '../_shared/emit-event.ts';
 import { checkProjectionReadback } from '../_shared/rpc-readback.ts';
 
 // Deployment version tracking
-const DEPLOY_VERSION = 'v16-deactivate-pattern-a-v2-readback';
+const DEPLOY_VERSION = 'v16-1-readback-schema-fix';
 
 // CORS headers for frontend requests
 const corsHeaders = standardCorsHeaders;
--- a/.claude/skills/infrastructure-guidelines/SKILL.md
+++ b/.claude/skills/infrastructure-guidelines/SKILL.md
@@ -445,6 +445,62 @@ $comment$Update user profile (first_name, last_name) via domain event
 
 ---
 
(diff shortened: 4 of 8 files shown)


infrastructure/supabase/supabase/functions/_shared/rpc-readback.ts
```diff
@@ -104,7 +104,18 @@ export async function checkProjectionReadback<T>(
...
   // surface the masked failure reason. Surface query errors verbatim
   // (RLS denials, missing GRANTs would be silent-failure modes if swallowed).
-  let readBackQuery = client.from(table).select('*').eq(lookupKey, lookupValue);
+  //
+  // Schema note: Edge Function service-role clients (e.g. `supabaseAdmin` in
+  // `manage-user/index.ts`) are typically constructed with `db: { schema: 'api' }`
+  // for RPC ergonomics. Projection tables and `domain_events` live in `public`,
+  // so the read-back must explicitly target the public schema via
+  // `.schema('public')`. Returned bug from first deployment 2026-05-12:
+  // "Could not find the table 'api.users' in the schema cache".
```
lars-tice, line 121: **P3 — non-blocking, address inline if cheap. Promote this invariant into the JSDoc / top-of-file docblock.**

This inline comment is the only place the schema-pinning invariant is documented. The function JSDoc (L64-94) and the top-of-file docblock (L1-46) are both silent on it. The next adopter of this helper (the queued `manage-user-reactivate-pattern-a-v2-retrofit/` card) will read the JSDoc first via IDE hover — not this inline comment buried inside the function body.

Suggest adding to the function JSDoc:

```
 * @remarks Hard-codes `.schema('public')` on every query because Pattern A v2
 * reads (projections + domain_events) MUST resolve in the public schema
 * regardless of how the caller constructed its client. Edge Function
 * service-role clients typically use `db: { schema: 'api' }` for RPC
 * ergonomics; without the explicit schema pin those clients resolve
 * `from('users')` as `api.users` (missing). See 2026-05-12 hotfix.
```

And a parallel section in the top-of-file docblock alongside the existing `# Transactional model` and `# PII masking` headings, e.g. `# Schema-pinning invariant`. The hotfix scope already paid the cost of writing the explanation; lifting it to the contract surface is ~5 lines.

---

**Rule-pinning cross-reference (added with the P1 elevation on `accept-invitation`)**

The exact wording of this docblock is now informed by the P1 in-PR action item: applying the same schema-pin fix to `accept-invitation/index.ts:515-519` and `:527-531`. The cumulative result of (a) this docblock, (b) the helper-side schema pins in PR #61, and (c) the `accept-invitation` fix will be the precedent set we use to decide whether to memorialize a SKILL/CLAUDE.md rule — specifically, whether the **"Edge Function admin clients constructed with `db: { schema: 'api' }` MUST schema-pin every `.from(...)` call to the table's owning schema"** invariant warrants pinning in `.claude/skills/infrastructure-guidelines/SKILL.md` and/or `infrastructure/supabase/CLAUDE.md` rather than living solely in per-file docblocks. The rule decision is **deferred until after the P1 fix lands** — see the top-level review body for the contingency statement. No rule is being written speculatively now; whatever shape the `accept-invitation` fix takes will inform whether the pattern is self-evident at the call site (no rule needed) or warrants a top-level discoverable rule (rule needed).

infrastructure/supabase/supabase/functions/_shared/rpc-readback.ts
```diff
@@ -104,7 +104,18 @@ export async function checkProjectionReadback<T>(
...
+  let readBackQuery = client
+    .schema('public')
+    .from(table)
+    .select('*')
+    .eq(lookupKey, lookupValue);
```
lars-tice, line 126: **P3 — already-imported local type. Optional cleanup.**

The helper defines a local `AnyClient = SupabaseClient<any, any, any>` at L57-58. `_shared/types.ts` already exports `AnySchemaSupabaseClient` with byte-identical shape (`_shared/types.ts:39`). Importing the shared one (already used by `manage-user/index.ts:28` and the rest of the Edge Function fleet) avoids a parallel-definition drift risk and lines up with the convention.

Not a blocker — the type is fine as-is — but the existing shared symbol is the canonical one for this purpose.

infrastructure/supabase/supabase/functions/manage-user/__tests__/deactivate-readback.test.ts
```diff
@@ -291,6 +307,43 @@ Deno.test('7. Helper does NOT swallow projection-table query errors', async () =
...
+
+  // Happy path issues exactly two queries (projection read-back + race-safe
+  // processing_error check). Both must explicitly route through public schema.
+  assertEquals(
+    schemaCalls.every((s) => s === 'public'),
+    true,
+    `every .schema() call must target 'public'; got: ${JSON.stringify(schemaCalls)}`,
+  );
+  assertEquals(
+    schemaCalls.length >= 2,
```
lars-tice: **P3 — strengthen the assertion. Catches a stronger class of regression.**

`schemaCalls.length >= 2` is correct for the happy path today (2 query sites: read-back + race-safe processing_error check). But the PR description explicitly calls out that a **third** missed `.schema('public')` was caught during test development — the race-safe lookup itself.

The current `>= 2` assertion would pass even if a future regression dropped the schema pin on ONE of the two remaining sites — as long as at least two sites kept theirs, the test stays green. Two ways to harden:

**Option A — exact count (preferred):** `assertEquals(schemaCalls.length, 2, ...)` for the happy path. Tells the next reader exactly how many query sites this helper is supposed to make; any addition or removal is a deliberate change that updates the assertion.

**Option B — invariant over `.from()`:** spy on every `.from()` call and assert each one was preceded by a `.schema('public')` (i.e. the chain shape). This is the actual invariant the helper promises — every `.from()` issues through `public`. Survives if future work adds a fourth query site, as long as the schema-pin discipline is maintained.

Option A is the smaller change; option B is a more durable contract test. Either closes the half-fix gap that the PR description acknowledges as the close-call during this PR's development.

infrastructure/supabase/supabase/functions/manage-user/index.ts
```diff
@@ -42,7 +42,7 @@ import { buildEventMetadata } from '../_shared/emit-event.ts';
 import { checkProjectionReadback } from '../_shared/rpc-readback.ts';
 
 // Deployment version tracking
-const DEPLOY_VERSION = 'v16-deactivate-pattern-a-v2-readback';
+const DEPLOY_VERSION = 'v16-1-readback-schema-fix';
```
lars-tice, line 45: **Verification — DEPLOY_VERSION is the sole version marker. Confirmed.**

Grepped the repo for stale `v16-deactivate-pattern-a-v2-readback` references. The only hits are inside `manage-user/index.ts` itself (this constant + the template strings that interpolate it). No AsyncAPI version field, migration `@a4c-rpc-shape` tag, or registry pin tracks the Edge Function deploy version. The patch-level `v16` → `v16-1-readback-schema-fix` bump is consistent with prior precedent (PR #32's `v11` series). Wire contract unchanged — `{success, error?, eventId}` envelope, HTTP 200 on handler failure, frontend caller `SupabaseUserCommandService.deactivateUser` (L305-364) parses `data.success` and surfaces `data.error` verbatim. No frontend changes needed.

lars-tice (commented):
## Verdict: APPROVE WITH IN-PR FIXES — 1 P1 + 3 P3, P1 required before merge

Targeted, well-tested hotfix for the helper-internal site. Three `.schema('public')` insertions are in the right places (every `.from(...)` callsite in the helper is now schema-pinned), the test addition is a genuine regression guard (would have caught the original PR #60 bug), and the wire contract is unchanged so the frontend caller (`SupabaseUserCommandService.deactivateUser`) needs no modification.

**However**, a same-bug-class instance exists outside the helper, in the `accept-invitation` Edge Function, and it has silent-failure semantics on the affected branch. Per the no-deferral discipline (small in-scope reviewer findings are addressed in the open PR, not punted to a card), it is elevated into this PR's action list. The action list now has one P1 (must land before merge) and three P3s (non-blocking).

### In-PR action list

**P1 — REQUIRED BEFORE MERGE. Apply the same schema-pinning fix to `accept-invitation/index.ts`.**

The same bug class PR #61 is fixing inside the helper has an external twin in `infrastructure/supabase/supabase/functions/accept-invitation/index.ts`:

- Admin client constructed with `db: { schema: 'api' }` at **L315-323**.
- Direct cross-schema reads:
  - **L515-519**: `supabase.from('users').select('deleted_at').eq('id', userId)` — `users` lives in `public`; currently resolves as `api.users` (missing) and PostgREST returns an error.
  - **L527-531**: `supabase.from('user_roles_projection').select('id').eq('user_id', userId)` — same problem with `user_roles_projection`.
- **L536-539** explicitly continues on error (`console.warn(...) // Continue - we'll emit user.created to be safe (idempotent via projections)`). **This is what makes it silent**: the affected branch always re-emits `user.created` instead of detecting an existing user. It is not actively producing a stack trace — it is silently mis-classifying every user the branch sees. This is the F1 silent-failure class Pattern A v2 was designed to retire, manifesting at the SDK boundary instead of the handler boundary.

**Fix shape options** (favored option in bold):

- **(a) Pin `.schema('public')` on the affected `.from()` calls at L515-519 and L527-531.** Mirrors PR #61's helper-side fix; localizes schema knowledge at the read site; one-line change per callsite. **Architecturally preferred** for consistency with how PR #61 closes the helper-side instance.
- (b) Construct a second admin client without the `db: { schema: 'api' }` override and use it for the cross-schema reads. Duplicates the SupabaseClient construction cost for a benefit option (a) already provides without the duplication. Acceptable but less clean than (a).

**Rule-pinning contingency (deferred until after the fix lands).** Once this fix is shaped and merged we will have three concrete examples of the same pattern (manage-user helper post-PR #61, accept-invitation post-fix, plus the JSDoc the P3 below adds to the helper). At that point the architect will revisit whether to memorialize an explicit rule in `.claude/skills/infrastructure-guidelines/SKILL.md` (and/or `infrastructure/supabase/CLAUDE.md`) stating that **Edge Function admin clients constructed with `db: { schema: 'api' }` MUST schema-pin every `.from(...)` call to the table's owning schema** (and likely an enumerated list of the canonical `public`-schema tables read across functions: `users`, `domain_events`, the `*_projection` tables). The rule decision is **deferred, not removed**: I am not writing a rule speculatively now because the in-PR fix shape may surface other affected sites I haven't yet enumerated, or it may make the pattern self-evident enough that an inline comment + docblock is the right resting place. The fix itself is the precedent that determines the right answer; the architect WILL revisit the rule decision after the in-PR fix lands.

**P3 — non-blocking.**

1. **Promote the schema-pinning invariant into the helper's function JSDoc + top-of-file docblock.** Inline comment at L108-113 is great but invisible to IDE hover and to the next adopter scanning the contract surface. Suggested wording in the inline comment thread on `_shared/rpc-readback.ts:113`.
2. **Strengthen the test-9 count assertion** from `>= 2` to `=== 2` (or add a `.from()` spy that asserts every `.from()` is preceded by a `.schema('public')`). Current assertion would pass a regression that drops the pin on one of the two remaining sites. The PR description's own narrative — a third missed call caught during test development — is the case in point.
3. **Reuse `AnySchemaSupabaseClient`** from `_shared/types.ts:39` instead of the locally-defined `AnyClient` alias. Avoids parallel-type drift; aligns with the rest of the Edge Function fleet.

### Architecture review checklist

| Item | Status | Note |
|---|---|---|
| CQRS Standards | PASS | Helper performs reads only; Pattern A v2 envelope contract preserved (success/failure shape). |
| Naming Conventions | PASS | `checkProjectionReadback`, `ProjectionReadbackResult<T>` — consistent with the codebase. |
| Design Patterns | PASS | Boundary helper / chokepoint pattern. Schema pinning is universal because Pattern A v2 reads ARE universally in `public` — hard-coded `.schema('public')` is the correct design, not a parameter. |
| data-testid Attributes | N/A | Backend code; no UI surface. |
| AsyncAPI Event Registration | PASS | No new events emitted; the helper only reads. Existing `user.deactivated` / `user.reactivated` events unchanged. |
| Type Generation — No Anonymous Types | PASS | `ProjectionReadbackResult<T>` is the discriminated union; success/failure paths are typed. Anonymous-shape callback args inside the test mock are acceptable per test-harness norms. |
| Observability, Tracing & Monitoring | NEEDS ATTENTION | Helper side is PASS — preserves correlation_id + tracing context; failures surface through the existing `console.error(...read-back failed...)` path with full event_id + masked error. **`accept-invitation` side fails this item** at L536-539 — the silent-continue on the cross-schema read masks every occurrence as a synthetic "new user" emission. P1 action item closes this. |
| Error Surfacing to UI | NEEDS ATTENTION | Helper side is PASS — returns `{success: false, error: masked}`; envelope packs it; frontend surfaces it on `UserVoidResult.error`. **`accept-invitation` side fails this item** at L536-539 — the silent-continue swallows the schema-mismatch error before it can reach the response envelope. P1 action item closes this. |

### Correctness audit

- **Three `.schema('public')` insertions in the helper are exhaustive** for the current implementation. Every `.from(...)` in `_shared/rpc-readback.ts` is now preceded by `.schema('public')`. Confirmed by reading the post-fix file end-to-end.
- **No fourth callsite** is naive in the helper. Exactly 2 query sites in the happy path and 1 additional in the NOT FOUND branch — all three are pinned.
- **Hard-coded `'public'` vs caller-config parameter** — the right call. Pattern A v2 by construction reads from projections + `domain_events`, both of which live in `public` schema regardless of how the caller's RPC client is configured. A parameter would shift the schema-correctness obligation onto every caller and reintroduce the exact silent-failure class this PR is closing. The trade-off cost (helper assumes `public` for read-back tables) is documented; should be promoted to the JSDoc per P3 item 1.
- **Reusability for `manage-user-reactivate-pattern-a-v2-retrofit/`** — unaffected. The schema pin is universal; reactivate will pass `expectedState: { is_active: true }` and the same `.schema('public')` resolution will work byte-equivalently.
- **Other Edge Functions with `db: { schema: 'api' }` admin clients** — audited and clean:
  - `validate-invitation/index.ts` — RPC-only, no direct `.from()`.
  - `invite-user/index.ts` (both client constructions at L531, L715) — RPC-only, no direct cross-schema `.from()` reads.
  - `manage-user/index.ts` — PR #61 fixes the issue via the helper; only direct `.from()` reads in this function go through the helper now.
  - **Only `accept-invitation/index.ts` requires the P1 fix.**

### Things considered raising but decided was overreach

- **"Add a CI lint that flags `db: { schema: 'api' }` + bare `.from('users' | 'domain_events' | ...projection)` patterns"** — buildable, but the rule shape and exception-list are non-trivial. Worth a dedicated piece of work if a fourth occurrence shows up after the P1 fix.
- **"Refactor manage-user to construct a separate `public`-schema client for the read-back, instead of relying on helper-side pinning"** — would localize the schema knowledge to the caller and arguably make the helper safer to misuse, but it duplicates the SupabaseClient construction cost (two clients per request) for a benefit the helper already provides. The current architecture (one client + helper-side schema pin) is fine.

### Lessons-learned postscript

This bug is a textbook **contract-implicit-from-config** gap: the helper's (and now `accept-invitation`'s) correct behavior depended on the caller's client configuration (`db.schema`) without any signal at the type level or the docblock that this dependency existed. Three things would have caught it earlier in PR #60 — and the same three would have caught the `accept-invitation` twin during its own original development:

1. **The helper's docblock** should have stated the schema-pinning invariant explicitly as a contract. The original PR #60 architect review could plausibly have asked "what schema does this helper resolve table names in, given that the only known caller has `db: { schema: 'api' }` configured?" — and didn't. P3 item 1 closes this gap on the helper side; the P1 fix on `accept-invitation` closes the parallel gap on that file.
2. **The test mock in PR #60 didn't model `.schema()`** at all — a hand-rolled mock with only `.from()` cannot catch a schema-pinning bug by construction. Test 9 added in this PR is the right shape; the assertion strengthening (P3 item 2) is what makes the regression class durable. Whatever test gets added alongside the `accept-invitation` fix should follow the same shape.
3. **Pattern A v2 was designed to close the F1 silent-failure class at the handler boundary**, but the wire-tier port introduced a new silent-failure class **at the SDK boundary** ("PostgREST returns row-not-found because the table-name-to-schema resolution silently picks the wrong schema"). The lesson — and worth memorializing alongside the existing F1 / `logIfPostgrestError` patterns — is that **every boundary helper (and every direct `.from()` call in an Edge Function with a schema-overridden client) should explicitly call out the schema/auth/grant context it assumes about its caller's client**. The same trap exists for `auth.admin.*` calls relying on service-role keys, for RLS-bypass implications of `db: { schema: 'api' }` clients, etc.

Memorialization timing: the user's memory already tracks F1-class silent-failure patterns. The right time to add an "Edge Function client-config invariants" entry (and the SKILL/CLAUDE.md rule decision) is **after the P1 fix on `accept-invitation` lands** — at that point we have three concrete examples (helper post-PR #61, accept-invitation post-fix, helper docblock) and can write the rule with real precedent rather than a speculative pattern. See the rule-pinning contingency under the P1 item above.

lars-tice:
**P1 — REQUIRED BEFORE MERGE. Same bug class as PR #61, external twin to the helper.**

> File anchor: `infrastructure/supabase/supabase/functions/accept-invitation/index.ts:539`
> (posted as a PR issue-comment because the file is outside PR #61's diff and the inline-anchor API rejects it; the in-PR action is to add this file to the PR's diff with the fix below)

PR #61 closes the schema-mismatch silent-failure inside `_shared/rpc-readback.ts` (helper-internal). The same bug class exists in `accept-invitation/index.ts` as an external twin, and **the branch at L536-539 is exactly what makes it silent**: the `console.warn` + comment "Continue - we'll emit user.created to be safe (idempotent via projections)" means the affected branch always re-emits `user.created` instead of detecting an existing user, never raising the schema-mismatch error to the response envelope.

**Trace** (all line refs in `infrastructure/supabase/supabase/functions/accept-invitation/index.ts`):

- **L315-323** — admin client constructed with `db: { schema: 'api' }`.
- **L515-519** — `supabase.from('users').select('deleted_at').eq('id', userId)` resolves as `api.users` (missing); `users` lives in `public`.
- **L527-531** — `supabase.from('user_roles_projection').select('id').eq('user_id', userId)` resolves as `api.user_roles_projection` (missing); the projection lives in `public`.
- **L536-539** — `if (rolesCheckError) { console.warn(...); /* Continue */ }` swallows the schema-mismatch error before it can reach the response envelope or the user. With `existingRoles === null` from the failed query, `isExistingUser === false` and the branch proceeds to re-emit `user.created` — exactly the silent mis-classification Pattern A v2 was designed to retire, manifesting here at the SDK boundary rather than the handler boundary.

**Fix shape options** (favored option in bold):

- **(a) Pin `.schema('public')` on the affected `.from()` calls at L515-519 and L527-531.** Mirrors PR #61's helper-side fix; localizes schema knowledge at the read site; one-line change per callsite. **Architecturally preferred** for consistency with how PR #61 closes the helper-side instance.
- (b) Construct a second admin client without the `db: { schema: 'api' }` override and use it for these cross-schema reads. Duplicates the SupabaseClient construction cost for a benefit option (a) already provides without the duplication. Acceptable but less clean than (a).

**Why this is in-scope for PR #61 rather than a follow-up card**: the silent-failure semantics at L536-539 make the bug class strictly worse here than in the helper-internal site that PR #61 already fixes — the helper at least raises the error through the existing envelope path. Per the no-deferral discipline, small in-scope reviewer findings are addressed in the open PR. There is no follow-up card.

**Rule-pinning contingency**: see the elevated P1 section in the top-level review body. The shape of this fix — combined with the helper-side fix from PR #61 and the docblock P3 on `_shared/rpc-readback.ts:113` — determines whether the architect adds a rule to `.claude/skills/infrastructure-guidelines/SKILL.md` (and/or `infrastructure/supabase/CLAUDE.md`) stating that **Edge Function admin clients constructed with `db: { schema: 'api' }` MUST schema-pin every `.from(...)` call to the table's owning schema**. The rule decision is **deferred until after this fix lands**, not removed entirely. No speculative rule is being written now; the fix itself is the precedent.

**Suggested test addition** (parallel to PR #61's test 9): a mock that fakes the `db: { schema: 'api' }` client behavior and asserts the existing-user-soft-deleted branch's `.from('users')` and `.from('user_roles_projection')` calls each issue against `public` rather than `api`. Strengthens the regression class for the whole `accept-invitation` file, not just this branch.

lars-tice (commented):
## Follow-up architectural verification — head SHA `4bf526da93a88c5472c7f925ec91d3935347b437`

Verifying the 8 findings from the prior review (1 P1 + 3 P3 + 4 lessons-learned) against commit `4bf526da93`. All findings line up with author's commit message intent though the author's F-numbering differs from the architect's — table below uses **architect numbering**.

### Verdict

**ALL 8 ADDRESSED — READY TO MERGE.** F1 (the P1 blocker) is closed by the chosen option (a) — explicit `.schema('public')` on every cross-schema `.from()` — extracted into a named `checkExistingUserPath` helper for testability. F2/F3/F4 (the P3 in-PR asks) all landed. F5/F6/F7 (lessons-learned) addressed materially. F8 (rule-pinning) recommendation below: **write the rule now** — three concrete precedents are visible and the surface area is bounded.

### Verification table

| # | Finding | Status | Evidence (file:line) |
|---|---------|--------|----------------------|
| **F1** | `accept-invitation` schema-pinning fix (P1) | ADDRESSED | `accept-invitation/index.ts:175-208` — extracted `checkExistingUserPath` helper; both queries pin `.schema('public')` at L181 + L194. No remaining `.from(<public-table>)` without schema pin (`grep` confirms exactly 2 `.from()` calls in the file, both in the helper, both preceded by `.schema('public')`). |
| **F2** | Schema-pinning invariant promoted to JSDoc + top-of-file docblock | ADDRESSED | `_shared/rpc-readback.ts:16-31` top-of-file `# Schema-pinning invariant` section with concrete PostgREST error text + reference to regression test pattern. Mirror docblock at `accept-invitation/index.ts:161-173` for the local helper. IDE-hover-visible at both adoption points. |
| **F3** | Replace local `AnyClient` alias with shared `AnySchemaSupabaseClient` | ADDRESSED | `_shared/rpc-readback.ts:66` imports `AnySchemaSupabaseClient` from `./types.ts`; `rpc-readback.ts:104` uses it as parameter type. Local alias gone. `grep -n "AnyClient\|AnySchemaSupabaseClient"` confirms single canonical source at `_shared/types.ts:39`. |
| **F4** | Strengthen `schemaCalls.length >= 2` to `.from()`-spy invariant | ADDRESSED (architect-preferred form) | `manage-user/__tests__/deactivate-readback.test.ts:328-374` Test 9 — `onFrom(table, currentSchema)` records every `.from()` call's schema scope; assertion loops `for (const call of fromCallLog) assertEquals(call.schema, 'public', ...)`. A regression that drops `.schema('public')` from any future code path would fire this with the offending table name. The count-`>=2` belt is preserved as a floor but is no longer load-bearing. |
| **F5** | Helper test mock models `.schema()` AND accept-invitation tests use the same pattern | ADDRESSED | Both test files use the identical mock-client construction (`makeMockClient` with `schemaImpl` returning `{from: (table) => fromImpl(table, schema)}`). Comparing `deactivate-readback.test.ts:74-101` vs `existing-user-check-schema.test.ts:48-83` — byte-equivalent invariant. The test harness now structurally cannot fail to model the schema scope. |
| **F6** | `accept-invitation` test coverage of the previously-silent failure mode | ADDRESSED | `existing-user-check-schema.test.ts:154-200` Test 5 is the F1 regression test. Six cases total: (1) new user happy path, (2) Sally scenario / existing user, (3) soft-deleted treated as new, (4) rolesCheckError surfacing, (5) schema-pinning invariant via `.from()`-spy, (6) deleted-user short-circuit. Test 5 fails by construction if either `.schema('public')` call is removed from the helper. Not a happy-path-only smoke. |
| **F7** | Visible vs swallowed error semantics on the `rolesCheckError` branch | ADDRESSED (with explicit choice) | `accept-invitation/index.ts:586-592` — the swallow remains but is now an **explicit documented choice**: schema-mismatch root cause is closed, the residual cause-set (RLS denial, missing GRANT, transient PostgREST) is logged via `console.warn` with the error object, and `// Continue - we'll emit user.created to be safe (idempotent via projections)` documents the fail-open rationale. The post-fix residual risk (mis-classifying an existing user under RLS denial) is downstream-observable rather than silent, and idempotent at the projection layer. Tightening to fail-closed would block legitimate accepts on transient PostgREST hiccups — author's choice is the right one. |
| **F8** | Rule-pinning decision | RECOMMEND **WRITE NOW** — see "Rule-pinning recommendation" below |

### Deno test results

```
deno test --allow-net manage-user/__tests__/ accept-invitation/__tests__/
ok | 26 passed | 0 failed (162ms)
```

Breakdown: 9 in `deactivate-readback.test.ts` (the original 8 + new Test 9 `.from()`-spy invariant), 6 in `existing-user-check-schema.test.ts`, 11 in `phone-id-resolution.test.ts` (pre-existing, unaffected).

### Rule-pinning recommendation (F8): **WRITE THE RULE NOW**

Three concrete precedents are now visible: (1) helper docblock at `_shared/rpc-readback.ts:16-31`, (2) extracted-helper docblock at `accept-invitation/index.ts:161-173`, (3) two regression-test patterns. The pattern is non-obvious enough to have bitten twice (in PR #60 helper rollout + the original `accept-invitation` code at L515-531 pre-extraction), and the surface area is bounded (only Edge Functions that mix `db: { schema: 'api' }` with `.from()` on a public-schema table). Counter-argument (CI lint as escalation path) is weaker here because: the bug manifests at runtime via PostgREST schema-cache miss, no static analyzer of the SDK chain catches it cheaply, and the audit cost of a future occurrence (silent `user.created` re-emit polluting audit trail) is higher than the cost of writing the rule now.

Neither target file is in the PR diff, so paste this in a follow-up commit:

**Target 1: `infrastructure/supabase/CLAUDE.md` § Critical Rules** (append after the existing read-back-pattern rule):

```markdown
> **⚠️ Edge Function clients with `db: { schema: 'api' }` MUST pin `.schema('public')` on every cross-schema `.from()`**
>
> Service-role clients in Edge Functions are typically constructed with
> `db: { schema: 'api' }` for RPC ergonomics (e.g., `supabaseAdmin` in
> `manage-user/index.ts`, `supabase` in `accept-invitation/index.ts`). This
> changes the default for **both** `.rpc()` AND `.from()` — but projection
> tables and `domain_events` live in `public`. Any naked `.from(<public-table>)`
> resolves as `api.<public-table>` and surfaces as:
>
>     Could not find the table 'api.<name>' in the schema cache
>
> Required form for every cross-schema read in an Edge Function:
>
> ```typescript
> const result = await client
>   .schema('public')   // ← required — every .from() needs its own
>   .from('users')
>   .select('id')
>   .eq('id', userId)
>   .maybeSingle();
> ```
>
> The bug class manifests at runtime (no compile-time signal), so the regression
> guard is the test invariant: every helper that does cross-schema reads MUST
> have a `.from()`-spy test that asserts every `.from()` is preceded by
> `.schema('public')`. Reference patterns:
> - `infrastructure/supabase/supabase/functions/_shared/rpc-readback.ts` (helper)
> - `infrastructure/supabase/supabase/functions/manage-user/__tests__/deactivate-readback.test.ts` Test 9 (invariant)
> - `infrastructure/supabase/supabase/functions/accept-invitation/__tests__/existing-user-check-schema.test.ts` Test 5 (invariant)
>
> History: introduced by PR #61 hotfix (2026-05-12) after two consecutive bug
> incidents — the helper's first deployment surfaced `api.users` schema-cache
> miss in UAT, and the `accept-invitation` audit found a pre-existing silent
> instance that was re-emitting `user.created` on every existing-user
> OAuth/SSO accept.
```

**Target 2: `.claude/skills/infrastructure-guidelines/SKILL.md`** — append as a new Rule (next available number) titled "Edge Function schema-pinning for cross-schema reads" with the same body condensed to skill-rule prose.

### Sanity checks performed

- F1 grep: `grep -n "\.from(" accept-invitation/index.ts` → exactly 2 hits (L182, L195), both in the extracted helper, both preceded by `.schema('public')`. No remaining unguarded sites.
- F5 audit (author's claim): `grep -rn "\.from(" .../functions/ --include="*.ts" | grep -v __tests__ | grep -v rpc-readback.ts | grep -v accept-invitation/index.ts` → zero hits. The "bug class contained to two files" claim is empirically true; `invite-user`, `organization-bootstrap`, `organization-delete`, `workflow-status` are pure-RPC.
- F2 docblock visibility: `_shared/rpc-readback.ts:16-31` and `accept-invitation/index.ts:161-173` both surface the invariant at IDE-hover for any future adopter.
- F4 invariant strength: a hypothetical regression that drops `.schema('public')` on the second query in the helper would fail Test 9 with `.from('domain_events') called without preceding .schema('public') — schema-pinning regression`. The assertion message names the offending table — actionable failure signal.

### Considered but did not raise

- **Tightening F7 to fail-closed**: Considered recommending that `rolesCheckError !== null` should be fatal post-schema-fix. Decided against — the residual cause-set (transient PostgREST / RLS / missing GRANT) includes legitimate transient failures, the fail-open behavior is idempotent at the projection layer, and the post-fix behavior is `console.warn`-observable. Fail-closed would trade a silent gap for a flaky-accept gap.
- **CI lint instead of doc rule** (F8 alternative): Considered recommending a regex-based grep CI guard `grep -n "supabase.from(" supabase/functions/*/index.ts` to catch naked cross-schema reads on functions that import a `db: { schema: 'api' }` client. Decided to write the doc rule first — the surface area is small enough that doc + test invariants suffice; if a third occurrence emerges, escalate to CI lint then.
- **Reactivate retrofit card pre-seeding the rule reference**: The next adopter is the reactivate retrofit, which lifts `checkProjectionReadback` unchanged — the test patterns and helper docblock will guide them. No card-level pre-seeding needed beyond the existing tasks.md note.

— Architect (Senior Fullstack Architect, follow-up verification)

PR #40, opened 2026-04-29
extract manage-user delete to api.delete_user + adjacent fixes

lars-tice:
## ⚠️ Merge instruction

**Do not squash-merge.** Preserve the eight migrations as separate commits to retain the forward-revert audit story (per architect review of PR #40). The history of M1→M2→M3→M4 followed by R1→R2→R3→R4 is intentional and load-bearing for future maintainers reading the migration trail.

## Summary

- Extracts `manage-user delete` operation into `api.delete_user` SQL RPC per [adr-edge-function-vs-sql-rpc.md](../blob/main/documentation/architecture/decisions/adr-edge-function-vs-sql-rpc.md). Inventory row #10 moves `candidate-for-extraction` → `extracted`.
- Bundles four adjacent fixes surfaced during the same work, all touching the user-management surface.
- Includes a same-day architectural course correction (scope-retrofit attempt → reverted) documented honestly in the commit history; see ADR Rollout 2026-04-27 § course correction.

## Commits (in order)

```
5eb2f864  fix(users): correct deactivate ban_duration; add reactivate unban path
d19acf49  refactor(users): remove self-bypass in update_user_notification_preferences
89db4ebb  docs(dev): file follow-up card for invitation-phone placeholder bug
fcaa2ed1  fix(users): add toast feedback to deactivate/reactivate/delete handlers
19b133ce  refactor(users): revert scope-permission retrofit; restore unscoped + seed design card
f89d3ba0  refactor(users): extract delete to api.delete_user + scope-permission retrofit
```

### Core extraction
- New RPC `api.delete_user(uuid, text)` — Pattern A v2 read-back against `users.deleted_at`, idempotency guard for already-deleted users, `access_blocked` JWT-claim guard, JWT-org_id-based tenancy guard returning same envelope as not-found (no UUID-existence leak).
- Edge Function `manage-user/index.ts`: `delete` removed from `Operation` union, perm switch, self-action guard, state guard, event-type switch, validation array, JSDoc. `DEPLOY_VERSION` → `v14-deactivate-ban-fix`.
- Frontend: `SupabaseUserCommandService.deleteUser` calls `api.delete_user` via `apiRpc`. 4 new envelope-contract tests added (success / handler-failure / 42501 / default reason).

### Architectural course correction (same day, reverted)

[...]

PR #63, opened 2026-05-12
feat(edge-functions): pivot deactivate + invitation-check to SQL RPC tier

lars-tice:
## Summary

Third (and final) attempt to retrofit Pattern A v2 read-back on `manage-user.deactivate`. **PR #60 (helper-side fix) and PR #61 (schema-pinning hotfix) BOTH failed UAT** with successive errors against the deployed PostgREST. This PR pivots both call sites from wire-tier to SQL-RPC tier, mirroring the `api.delete_user` precedent (PR #40).

### Why the previous attempts failed

The deployed Supabase project's PostgREST exposes ONLY the `api` schema. Both wire-tier approaches failed at runtime:

| Attempt | Approach | UAT Error |
|---|---|---|
| PR #60 | Naked `.from('users')` with `db: { schema: 'api' }` client | `Could not find the table 'api.users' in the schema cache` |
| PR #61 | `.schema('public').from('users')` chain | `The schema must be one of the following: api` |

PostgREST rejects schema switches to non-allowlisted schemas at the gateway. The local `config.toml:14` `schemas = ["public","graphql_public","api"]` does not reflect the deployed runtime config.

### The pivot

Move the Pattern A v2 read-back from Deno (Edge Function tier) to SQL (PG tier). The Edge Function stays — LB1 still requires `auth.admin.updateUserById` for the ban — but the emit + projection-verification portion is now an `api.deactivate_user` RPC call. RPC bodies have full SQL access to all schemas regardless of PostgREST's exposed-schemas allowlist; PostgREST only needs to find the RPC entry point in `api`. Mirrors the precedent `api.delete_user` (PR #40) shape byte-for-byte.

### What changed

- **Migration** `20260512194836_deactivate_user_rpc_and_check_user_invitation_existence.sql`:
  - `api.deactivate_user(p_user_id uuid, p_reason text DEFAULT NULL) RETURNS jsonb` — envelope; pre-emit guards (auth, access_blocked, `has_permission('user.update')`), tenancy guard via JWT org_id + `users.current_organization_id` lookup, idempotency (already-inactive, already-deleted), emit `user.deactivated`, Pattern A v2 read-back (BOTH checks per Rule 13).
  - `api.check_user_invitation_existence(p_user_id uuid) RETURNS jsonb` — read; replaces the wire-tier `.from('users')` / `.from('user_roles_projection')` reads in `accept-invitation`. Soft-deleted users treated as NEW (re-invitation flow).
  - Both tagged `@a4c-rpc-shape: envelope|read` per Rule 17.

- **Edge Function `manage-user` (v17)**: deactivate path is now a thin RPC orchestrator. Reactivate path unchanged (retrofit deferred to its own card).

- **Edge Function `accept-invitation` (v24)**: `checkExistingUserPath` body rewritten to call the RPC. Function signature preserved so call sites are untouched.

- **Deleted** `_shared/rpc-readback.ts` (architecturally infeasible).

- **Tests**:
  - Deleted `manage-user/__tests__/deactivate-readback.test.ts` (tested the deleted helper).
  - Rewrote `accept-invitation/__tests__/existing-user-check-schema.test.ts` — 6 cases, all passing locally.

- **Rule 19 INVALIDATED** in `.claude/skills/infrastructure-guidelines/SKILL.md` + `infrastructure/supabase/CLAUDE.md`:
  - PR #61 rule body ("MUST pin `.schema('public')` on every cross-schema `.from()`") replaced with
  - "Edge Functions MUST NOT use PostgREST for cross-schema reads — use `api.*` SQL RPC instead"
  - PR #60 / PR #61 history preserved as cautionary lesson.

- **Frontend types**: regenerated `database.types.ts` (frontend + workflows, byte-identical) and patched `rpc-registry.generated.ts` for the two new RPCs.

## Test plan

- [x] `supabase db push --linked` — migration applied to dev DB
- [x] `supabase gen types typescript --linked` — types regenerated for both consumers
- [x] `npm run gen:rpc-registry` (equivalent — patched manually since local container not running; CI workflow `rpc-registry-sync.yml` will validate against fresh container)
- [x] `cd frontend && npm run typecheck` — green
- [x] `cd workflows && npm run typecheck` — green
- [x] `cd frontend && npm run lint -- --max-warnings 0` — green
- [x] `cd frontend && npm run build` — green
- [x] `cd frontend && npm run test -- --run` — pre-existing 56-fail baseline unchanged
- [x] `deno test --allow-net manage-user/ accept-invitation/ _shared/` — 96/96 pass
- [ ] **UAT against dev** (post-deploy) — required merge gate:
  - **Test 1** (happy path deactivate): Lars deactivates a non-self user in the org; expect `{success: true, eventId: ...}`; UI shows user as inactive; `auth.users.ban_duration` is set; user cannot log in.
  - **Test 2** (Pattern A v2 failure surfacing): force handler to fail (e.g., revoke handler GRANT or simulate trigger error); expect `{success: false, error: 'Event processing failed: ...'}` from envelope; UI surfaces the error without claiming success.
  - **Test 3** (idempotency — already deactivated): re-run deactivate on the now-inactive user; expect `{success: false, error: 'User is already deactivated'}` envelope.
  - **Test 4** (cross-tenant tenancy): attempt to deactivate a userId from another tenant (constructed); expect `{success: false, error: 'User not found in this organization'}` — same envelope as not-found (no UUID-existence leak).
  - **Test 5** (Sally scenario, accept-invitation): existing user (already has role in another org) accepts invitation to a new org via OAuth/SSO; expect `user.created` is NOT re-emitted (verify via `domain_events` query); user gets the new role.
  - **Test 6** (re-invitation of soft-deleted user): soft-delete a user, then invite + accept again with same email; expect full `user.created` flow runs (isDeleted=true → treat as NEW).
  - **Test 7** (reactivate regression check): reactivate the deactivated user; expect existing legacy emit flow still works (unchanged by this PR).
  - **Test 8** (partial failure — RPC succeeds, ban fails; added per F2 architect-review finding): deactivate a target user via the UI. Before the request fires, temporarily revoke `auth.admin.updateUserById` for the service-role key (e.g. via a brief `auth.users` RLS policy that denies UPDATE), OR force a `400` from the Supabase Auth API via a malformed `ban_duration` in a dev-only branch. Expected: frontend sees `{success: true, userId, operation: 'deactivate', eventId}`; backend logs at WARN level `[manage-user v<ver>] Failed to ban auth user: <error>`; `domain_events` row for the `user.deactivated` event has `processing_error IS NULL`; `public.users.is_active = false` for the target; `auth.users.banned_until` for the target is UNCHANGED. **Note**: this documents the intentional fail-open posture for ban failures — the projection state is correct; the ban is recoverable via a re-run of deactivate or via a future reconcile-ban admin operation. If product later decides to fail-closed on ban failures, this scenario becomes the test point.

## Notes

- **Single PR confirmed** with the user during plan approval (deactivate-pivot + accept-invitation-pivot + Rule 19 invalidation + helper deletion all together).
- **Pre-deploy ritual gap**: this is the second consecutive case where local Deno tests passed but deployed UAT failed. Rule 19's restated form codifies the lesson: smoke against real Supabase BEFORE opening PR. The card's `tasks.md` now carries this discipline forward to the next Edge Function card (deactivate-reactivate retrofit).


infrastructure/supabase/supabase/migrations/20260512194836_deactivate_user_rpc_and_check_user_invitation_existence.sql
```diff
@@ -0,0 +1,269 @@
...
+
+    RETURN jsonb_build_object(
+        'success', true,
+        'eventId', v_event_id,
+        'userId', p_user_id
+    );
+END;
+$$;
+
+GRANT EXECUTE ON FUNCTION api.deactivate_user(uuid, text) TO authenticated;
```
lars-tice, line 200: **NT-1 (P3, doc/operability)**: `api.deactivate_user` is `GRANT EXECUTE … TO authenticated`-only, while `api.check_user_invitation_existence` (L251-252) grants to BOTH `authenticated` and `service_role`. The reason is right (the existing-user check is called from an Edge Function whose admin client uses service_role), but the asymmetric grant is non-obvious enough that a future reader could mis-copy. Consider adding a one-line comment on each GRANT explaining the audience — e.g. above L190 `-- authenticated only; called by manage-user Edge Function with the caller's JWT` and above L251-252 `-- both: authenticated for direct calls, service_role for accept-invitation admin client path`. No code change required.

infrastructure/supabase/supabase/migrations/20260512194836_deactivate_user_rpc_and_check_user_invitation_existence.sql
```diff
@@ -0,0 +1,269 @@
...
+    WHERE id = p_user_id AND is_active = false;
+
+    IF NOT FOUND THEN
+        SELECT processing_error INTO v_processing_error
+        FROM public.domain_events WHERE id = v_event_id;
+        RETURN jsonb_build_object(
+            'success', false,
+            'error', 'Event processing failed: ' ||
+                COALESCE(v_processing_error, 'projection read-back returned no row'),
+            'eventId', v_event_id
```
lars-tice, line 175: **NT-2 (P3, asymmetry with precedent — verify intentional)**: The success envelope here includes `'eventId', v_event_id` (also on the success path at L184). `api.delete_user` (PR #40, migration `20260427205333_extract_delete_user_rpc.sql` L118-122) does NOT include `eventId` on the success-false read-back miss path, only on the success path (L141). The Edge Function caller (`manage-user/index.ts:331`) explicitly surfaces `env_.eventId` to the frontend response on the failure path — good for audit-log deep-linking. This is a reasonable strict-superset of the precedent, but worth confirming in the function COMMENT that the eventId on failure envelopes is intentional (the docstring on `api.delete_user` documents an envelope that does NOT carry eventId on failure). If kept, consider extending the COMMENT on L192-200 to call out `eventId` on the success-false read-back paths as part of the contract — protects against a future "normalize to match delete_user" cleanup PR that would silently drop the field.

infrastructure/supabase/supabase/migrations/20260512194836_deactivate_user_rpc_and_check_user_invitation_existence.sql
```diff
@@ -0,0 +1,269 @@
...
+    v_event_id uuid;
+    v_processing_error text;
+    v_now timestamptz := now();
+BEGIN
+    -- =====================================================================
+    -- PRE-EMIT GUARDS (RAISE EXCEPTION; no audit row yet)
+    -- =====================================================================
+
+    IF v_caller_id IS NULL OR v_org_id IS NULL THEN
+        RAISE EXCEPTION 'Access denied' USING ERRCODE = '42501';
```
lars-tice, line 78: **NT-3 (P3, defense-in-depth consideration — not a blocker)**: The pre-emit `has_permission('user.update')` guard is unscoped (correct per memory's user-model rule: A4C users have no organizational location finer than tenant). The Edge Function (`manage-user/index.ts:218`) ALREADY performed an in-Deno `hasPermission(effectivePermissions, 'user.update')` check on the JWT claims. Both checks fire on every deactivate call — the SQL one is now the authoritative one and the Deno one is belt-and-suspenders. Worth a brief comment in the migration header noting this is intentional (both layers gate the same permission); otherwise a future Edge Function cleanup PR might delete the Deno-side check on grounds of "the RPC enforces it" and we'd lose the early-rejection path (saves an RPC round-trip on permission failures). Up to author whether to capture this in code or memorialize in the followup card.

infrastructure/supabase/supabase/functions/manage-user/index.ts
```diff
@@ -277,139 +272,129 @@ serve(async (req) => {
...
-          user_id: user.id,
-          reason: requestData.reason || `Manual ${requestData.operation}`,
-        }),
+    if (requestData.operation === 'deactivate') {
+      console.log(`[manage-user v${DEPLOY_VERSION}] Calling api.deactivate_user RPC...`);
+
+      // RPC must use caller's JWT for `auth.uid()` + `org_id` claim + permission
+      // gate. Build a per-call client with the user's auth header instead of the
+      // service-role client above.
+      const supabaseRpcCaller = createClient(env.SUPABASE_URL, resolveAnonKey(req, env), {
```
lars-tice: **F1 (P2, code quality, optional)**: `supabaseRpcCaller` is constructed inline inside the deactivate branch (L300-303) — a third Supabase client alongside `supabaseUser` (L185) and `supabaseAdmin` (L265). The reason is correct (need caller JWT + `db: { schema: 'api' }` for the RPC), but the three-client surface is non-obvious. Two small consolidations are possible: (a) hoist the per-request `supabaseRpcCaller` construction up next to `supabaseUser` so all clients are declared in one block with a comment explaining when to use which, or (b) reuse `supabaseUser` with a per-call `.schema('api').rpc(...)` chain — that approach was rejected for `accept-invitation` because cross-schema chains hit the gateway, but `.schema('api').rpc(...)` on the api-allowlisted PostgREST is exactly the case that DOES work. Not a blocker — just a future-readability nit.

infrastructure/supabase/supabase/functions/manage-user/index.ts
```diff
@@ -433,21 +415,18 @@ serve(async (req) => {
 
       if (banError) {
         console.warn(`[manage-user v${DEPLOY_VERSION}] Failed to ban auth user:`, banError);
-        // Don't fail the request - event was emitted, projection will update.
-        // But surface this as a warning since the auth-tier ban is the actual
-        // login-prevention mechanism; without it the user can still log in.
+        // Don't fail the request - projection update already succeeded via RPC.
+        // The auth-tier ban is the actual login-prevention mechanism; without
+        // it the user can still log in. Surface as a warning for ops visibility.
       }
```
lars-tice, line 420: **F2 (P2, partial-failure visibility)**: After `api.deactivate_user` succeeds, `auth.admin.updateUserById({ban_duration: '876000h'})` runs (L411-414). If THIS call fails, the function continues, swallows the error as a `console.warn` (L417), and returns `success: true` to the frontend (L436-441) — but the user can still log in. This is documented in the comment block at L399-408 ("don't fail the request — projection update already succeeded"), and the comment is defensible (projection state is right, ban is the actual login-prevention mechanism). But the PR description's Test 2 ("Pattern A v2 failure surfacing") doesn't exercise THIS partial-failure path — the auth-ban failure mode after RPC success. Two options: (i) add a UAT test case for "deactivate succeeds at projection tier but auth ban fails — assert UI sees a warning surface", OR (ii) explicitly fail-closed: return success=false with a different error code when the ban fails (cleaner contract, projection rolls back via compensating reactivate? — but reactivate emits an event so that's not free). Up to author — calling out so it's seen and ruled on, not blocking.

.claude/skills/infrastructure-guidelines/SKILL.md
```diff
@@ -445,59 +445,56 @@ $comment$Update user profile (first_name, last_name) via domain event
...
 
-// In the test:
-for (const call of fromCallLog) {
-  assertEquals(call.schema, 'public',
-    `.from('${call.table}') called without preceding .schema('public') — schema-pinning regression`);
-}
+```bash
+# Find any wire-tier read against a non-api table — every match needs review.
+grep -rnE "\.from\('([^']+)'\)" infrastructure/supabase/supabase/functions/ \
+  | grep -vE "_shared/__tests__|__tests__/.*\.test\.ts"
```
lars-tice, line 505: **F3 (P2, Rule 19 wording — carve-out clarity)**: The new Rule 19 says "Edge Functions MUST NOT use PostgREST for cross-schema reads — use `api.*` SQL RPC instead." That's the right architectural rule. But the wording could be misread as banning `db: { schema: 'api' }` clients entirely. Concretely: `accept-invitation/index.ts:387-395` and `manage-user/index.ts:265-273` both still use `db: { schema: 'api' }` clients and these are LEGITIMATE — they only do `.rpc()` calls. The audit query `grep -rnE "\.from\('([^']+)'\)" .../functions/` will return zero hits in those files now, so the audit catches the actual bug class. Worth adding one explicit sentence right after the corrected/wrong code block: "`db: { schema: 'api' }` clients are still the correct construction for Edge Functions that exclusively call `.rpc()` — Rule 19 only governs `.from()` calls." Otherwise a future reader running through the rule and seeing `db: { schema: 'api' }` on the admin client could conclude the construction itself is the bug. The CLAUDE.md mirror at `infrastructure/supabase/CLAUDE.md` has the same wording and would benefit from the same one-line carve-out.

infrastructure/supabase/supabase/functions/accept-invitation/__tests__/existing-user-check-schema.test.ts
```diff
@@ -106,121 +88,102 @@ Deno.test('1. New user (no roles, not deleted): isExistingUser=false', async ()
...
+      rpcCalls.push({ name, args });
+    },
+  });
+
+  await checkExistingUserPath(client, USER_ID);
 
+  assertEquals(rpcCalls.length, 1, 'helper must invoke exactly one RPC');
+  assertEquals(rpcCalls[0].name, 'check_user_invitation_existence');
+  assertEquals(rpcCalls[0].args, { p_user_id: USER_ID });
+});
```
lars-tice, line 197: **F4 (P3, test rewrite — invariant assertion strength)**: This test is the new architectural-invariant guard, replacing the old `.from()`-spy invariant from PR #61. The current implementation asserts (a) exactly one RPC call, (b) it's named `check_user_invitation_existence`, (c) args are `{p_user_id: USER_ID}`. The doc comment (L150-154) explains the IMPLICIT complement guard: if a refactor calls `client.from(...)` the mock throws because the mock doesn't define `from`. That's a clever and reasonable falsification path, but it's brittle (a future test author who adds `.from` to the mock client for some other test case may not realize they're disabling the guard). Consider one of: (i) make the absence-of-`.from` test explicit — add a separate assertion or test case that defines `.from()` on the mock as a throw-stub with a message like "Test invariant violated: helper made a `.from()` call; Rule 19 prohibits wire-tier cross-schema reads", and assert that throw-stub is never invoked. (ii) Or skip the explicit test and rely on the SKILL.md audit query running in CI (which would need to be wired). Either is fine — current state is adequate but not as load-bearing as the doc claims.

lars-tice (commented):
## Verdict: APPROVE WITH IN-PR FIXES

Third-time-correct on this load-bearing feature. The architectural pivot from wire-tier (PR #60 / PR #61) to PG-tier (this PR) is the right move and mirrors the established `api.delete_user` precedent (PR #40) shape-for-shape. The diagnosis is sound, the implementation is clean, and the rule update correctly memorializes the bug class with PR #60/#61 preserved as cautionary lesson. Nothing here blocks merge — five findings (3 NTs + 2 Fs), all non-blocking.

## Verification performed

- Migration `20260512194836_…` read in full. Mirrors `api.delete_user` precedent at `infrastructure/supabase/supabase/migrations/20260427205333_extract_delete_user_rpc.sql`: SECURITY DEFINER, search_path lockdown, pre-emit guards with `RAISE EXCEPTION USING ERRCODE`, idempotency envelope returns (not RAISE), Pattern A v2 read-back with BOTH checks (Rule 13).
- Rule 16 audit on the migration: all 3 `RAISE EXCEPTION` strings are literal (no `%` identifier interpolation). Clean.
- Rule 17: both functions tagged `@a4c-rpc-shape: envelope|read` via `COMMENT ON FUNCTION` in the SAME migration. CREATE OR REPLACE preserves OID, but neither function pre-existed in baseline_v4 (audited) so this is clean creation, no DROP+CREATE re-tag concern.
- Rule 15 baseline-overload: `grep "FUNCTION \"api\"\.\"deactivate_user\"" baseline_v4.sql` → zero hits. Same for `check_user_invitation_existence`. Clean.
- `frontend/src/types/database.types.ts` and `workflows/src/types/database.types.ts` are byte-identical: `md5sum` produces the same `ae2ecb1f5d66d3e28fe8042b981eed0d` hash for both. Both contain the new RPC signatures.
- `rpc-registry.generated.ts`: `'deactivate_user'` added to `EnvelopeRpcs`, `'check_user_invitation_existence'` added to `ReadRpcs`. Classification correct per the COMMENTs.
- Deleted helper: `_shared/rpc-readback.ts` is GONE. Deleted test: `manage-user/__tests__/deactivate-readback.test.ts` is GONE (the entire `__tests__` directory under `manage-user/` is gone). No residual imports — `grep -rn rpc-readback infrastructure/supabase/supabase/functions/` returns zero hits in source files (only doc-string history references remain in the new code).
- `AnySchemaSupabaseClient` continues to live in `_shared/types.ts` (not the deleted helper). All remaining imports point at the right file. Clean.
- Audit query `grep -rnE "\.from\('([^']+)'\)" infrastructure/supabase/supabase/functions/ | grep -vE "__tests__|\.test\.ts"` → only matches are in code comments (docstring history in `accept-invitation/index.ts`). Zero live wire-tier `.from()` calls.
- Deno tests: 96 passed | 0 failed (424ms). Confirmed locally.
- Tenancy column: `users.current_organization_id` is consistent with the schema's user-as-identity model (per memory) and with `handle_user_created` / `api.get_user_by_id` precedent. Correctly NOT `current_org_unit_id` (per-shift session state).
- Reactivate path correctly untouched — retrofit deferred to next card. Memory shows the deactivate-retrofit card directory exists (`dev/active/manage-user-deactivate-pattern-a-v2-retrofit/`); the reactivate-retrofit card is acknowledged in memory as "NOT YET SEEDED" — not blocking this PR.

## Findings

**P3 — non-blocking, address in this PR if convenient or take to followup:**

- **NT-1** (`migration:190`): GRANT asymmetry between the two new RPCs (one `authenticated`-only, the other both `authenticated` + `service_role`). Reason is correct, but one-line comments on each GRANT would prevent mis-copy.
- **NT-2** (`migration:167`): success-false envelope includes `eventId` on read-back miss path; `api.delete_user` precedent does NOT. Reasonable strict superset (Edge Function caller surfaces it for audit-log deep-linking), but worth memorializing in the function COMMENT to protect against a future "normalize to match delete_user" cleanup.
- **NT-3** (`migration:78`): SQL `has_permission('user.update')` guard duplicates the in-Deno `hasPermission(..., 'user.update')` check on the Edge Function caller. Both fire on every call. Intentional (RPC is authoritative; Deno is belt-and-suspenders early-rejection). Worth a note in the migration header so a future cleanup doesn't strip the Deno check.

**P2 — non-blocking but visible to operators:**

- **F1** (`manage-user/index.ts:300`): three-Supabase-client surface (`supabaseUser`, `supabaseAdmin`, inline `supabaseRpcCaller`). Optional consolidation hoists the per-request RPC caller construction up next to `supabaseUser` for one-block visibility.
- **F2** (`manage-user/index.ts:421`): post-RPC `auth.admin.updateUserById` failure path returns `success: true` with only a `console.warn`. Documented but not exercised by UAT Test 2. Suggest extending Test 2 to cover this partial-failure path OR explicitly fail-closed.
- **F3** (`.claude/skills/infrastructure-guidelines/SKILL.md:485`): Rule 19 reads as banning `db: { schema: 'api' }` clients entirely; it's not — the rule only governs `.from()`. One sentence of carve-out language would prevent future misreadings. Same edit needed in the CLAUDE.md mirror.
- **F4** (`existing-user-check-schema.test.ts:172`): the architectural-invariant guard relies on an IMPLICIT "mock client doesn't define `.from`, so refactoring back to `.from()` makes the test crash" property. Clever but brittle — a future test author could disable the guard unwittingly. Suggest making the absence assertion explicit via a throw-stub.

## Architecture concerns surfaced by the pivot

- **Local vs deployed config drift class.** `supabase/config.toml:14` lists `schemas = ["public","graphql_public","api"]` but the deployed PostgREST exposes only `api`. This config divergence is now a documented bug class (Rule 19 history captures it well), but the pre-deploy ritual is the only thing closing it: "smoke against real Supabase BEFORE opening PR." The PR description claims `tasks.md` carries this discipline forward — **the `tasks.md` diff is uncommitted in the working tree** (I observed `modified: dev/active/manage-user-deactivate-pattern-a-v2-retrofit/tasks.md` in `git status` after checking out the branch). The Phase 7 SQL-RPC pivot section is good content. Recommend committing it before merge — it's the only place the discipline carries forward and it's a non-zero-cost piece of organizational memory.
- **Permission-gate redundancy.** SQL `has_permission` + Edge Function `hasPermission` both gate the same permission. This is current convention across the codebase (PR #36, #39, #40 also kept the Edge Function gate post-extraction). Worth memorializing as a pattern in the next ADR revision rather than as a per-PR comment.

## Considered raising but decided was overreach

- **`v${DEPLOY_VERSION}` double-`v` cosmetic bug** in `accept-invitation/index.ts:30` and `manage-user/index.ts:52` — the `DEPLOY_VERSION = 'v24-...'` constant is interpolated as `` v${DEPLOY_VERSION} `` producing `vv24` in logs. Confirmed pre-existing (main branch has `v23-...` and `v16-1-...` with the same pattern). Not this PR's regression and not a behavior bug.
- **Asking for an integration test exercising real Supabase PostgREST** — the right answer is the UAT test plan in the PR description, not unit-style mocks. Author has already designed a 7-scenario UAT plan. Pushing for more pre-merge test infra would be overreach.

## Architecture Review Checklist

- **CQRS Standards** — PASS. Both RPCs follow the established Pattern A v2 envelope contract for writes / read-shape for reads. `api.deactivate_user` emits `user.deactivated` (no direct projection writes). `api.check_user_invitation_existence` reads from existing projections.
- **Naming Conventions** — PASS. `api.deactivate_user` mirrors `api.delete_user`. `api.check_user_invitation_existence` is descriptive. Parameters use `p_*` convention. Internal vars use `v_*` convention. Function COMMENTs follow established docstring shape.
- **Design Patterns** — PASS. Pattern A v2 (Rule 13 BOTH checks) correctly applied. Tenancy guard uses canonical JWT-org_id + `users.current_organization_id` pattern. Idempotency uses success-false envelope (not RAISE), preserving the success path for the auth-tier ban call.
- **data-testid Attributes** — N/A (no frontend UI changes).
- **AsyncAPI Event Registration** — PASS. `user.deactivated` is a pre-existing event type — no AsyncAPI surface area added.
- **Type Generation — No Anonymous Types** — PASS. `database.types.ts` regenerated for both consumers (byte-identical MD5). `rpc-registry.generated.ts` updated. Edge Function uses typed inline shape for envelope but accesses through narrowed fields, not anonymous structural typing.
- **Observability, Tracing & Monitoring** — PASS. `domain_events.processing_error` is the diagnostic path; the RPC surfaces it through the envelope. Edge Function preserves `correlationId` + `tracingContext` + `createSpan`/`endSpan`. Structured logs with `[manage-user v${DEPLOY_VERSION}]` prefix retained.
- **Error Surfacing to UI** — PASS. Envelope `error` string propagates from SQL → Edge Function (L327-336) → frontend response. Success-false path returns HTTP 200 with `success: false` + typed error, per Pattern A v2 contract. Frontend caller parses `data.success` not HTTP status, which is the documented convention.

The five P2/P3 findings are tractable in this PR (the largest is F3 — one carve-out sentence). Default verdict per current guidance: APPROVE WITH IN-PR FIXES.

lars-tice (commented):
## Architect follow-up review — verdict: **ALL 10 ADDRESSED — READY TO MERGE** (pending UAT)

Re-review of commit `1568c768c3` against the 7 inline findings + 3 architecture concerns from the prior architect review. All ten items now show concrete fixes in the diff; no remaining gaps and no regressions surfaced.

### Verification table

| # | Item | Status | Evidence (file:line) |
|---|---|---|---|
| F1 | 3-client surface (`supabaseUser`/`supabaseAdmin`/`supabaseRpcCaller`) | **ADDRESSED** — `supabaseRpcCaller` eliminated; deactivate path now reuses `supabaseUser` via per-call `.schema('api').rpc()`. Better than relocation — one fewer client. | `manage-user/index.ts:302-307` |
| F2 | Post-RPC `auth.admin.updateUserById` failure returns `success: true` with only `console.warn` | **ADDRESSED (option c)** — UAT scenario 8 (partial-failure ban) added to PR body with concrete instrumentation procedure + an explicit "intentional fail-open posture" justification. The product decision (projection-state-is-correct; ban is recoverable) is now documented and testable rather than incidental. | PR body Test 8; `manage-user/index.ts:415-420` (existing warn block) |
| F3 | Rule 19 carve-out for `db:{schema:'api'}` clients | **ADDRESSED** — carve-out paragraph + two-pattern code block + "client construction is incidental" invariant statement now present in both files. | `.claude/skills/infrastructure-guidelines/SKILL.md:476-494`; `infrastructure/supabase/CLAUDE.md:324-344` |
| F4 | Implicit absent-`.from` invariant in Test 5 | **ADDRESSED** — explicit throw-stub on `.from()` plus `__fromCalls`-length assertion. Failure message points future authors directly at Rule 19. | `existing-user-check-schema.test.ts:70-83` (stub) + `:192-196` (assertion) |
| NT-1 | GRANT-audience asymmetry | **ADDRESSED** — one-line audience comments on both RPCs explain the authenticated-only vs authenticated+service_role split. | migration `:198-199` (deactivate_user); `:273-276` (check_user_invitation_existence) |
| NT-2 | `eventId` on success-false envelope | **ADDRESSED** — `COMMENT ON FUNCTION api.deactivate_user` now explicitly documents the envelope contract including the "load-bearing, not vestigial" rationale that prevents a future normalize-to-`api.delete_user` cleanup from dropping the field. | migration `:207-217` |
| NT-3 | SQL/Deno `user.update` duplication | **ADDRESSED** — migration body now carries an "INTENTIONAL DUPLICATION" block stating the SQL guard is authoritative and the Deno guard is belt-and-suspenders early-rejection. | migration `:91-97` |
| Arch-1 | `config.toml` local-vs-deploy drift | **ADDRESSED** — schemas list tightened from `["public","graphql_public","api"]` to `["api"]`. This closes the runtime gap: local PostgREST now mirrors the deployed allowlist, so any future wire-tier `.schema('public').from(...)` attempt fails locally with the same gateway error that bit PR #61 in deployed UAT. Side-effect audit confirmed in commit message (psql-based rpc-registry-sync CI unaffected; zero GraphQL usage; frontend CQRS+ESLint already forbid wire-tier reads). Note: `extra_search_path = ["public","extensions"]` remains intact so RPC bodies can still reference public tables natively in SQL — only the wire-tier exposure tightens. | `config.toml:26` (and explanatory comment `:14-25`) |
| Arch-2 | Uncommitted `tasks.md` discipline doc | **ADDRESSED** — Phase 7 SQL-RPC-pivot ledger committed (10 [x] line items capturing the trail) + a "Process discipline (template)" section codifying the pre-deploy smoke-against-real-Supabase rule with explicit "this discipline was first formalized DURING this card's hotfix cycle" framing. Reactivate retrofit (next card) is explicitly bound to satisfy this before opening its PR. | `dev/active/manage-user-deactivate-pattern-a-v2-retrofit/tasks.md:60-77` (Phase 7) + `:81-85` (Process discipline) |
| Concern #3 | Permission-gate redundancy as codebase-wide pattern | **N/A — correctly not pursued** — observation only; no diff inflation. |

### Test verification

- **Deno tests**: 96/96 pass (`deno test --allow-net manage-user/ accept-invitation/ _shared/`).
- **Wire-tier `.from()` audit**: `grep -rnE "\.from\('([^']+)'\)" infrastructure/supabase/supabase/functions/ | grep -vE "__tests__|\.test\.ts"` returns only documentation-comment matches in `accept-invitation/index.ts:163,166` — zero actual `.from()` invocations remain.

### Concern #1 specifics — does the config.toml change actually close the gap?

Yes. The author chose to tighten (`schemas = ["api"]`) rather than document-the-superset-with-a-comment. This is the strong fix:

1. Local PostgREST now refuses cross-schema `.from(public.X)` reads at the gateway, producing the same error class that surfaced in deployed PR #60/#61 UAT.
2. The next Edge Function author writing a Deno integration test that runs against `supabase start` will hit the failure locally rather than discovering it in UAT.
3. `extra_search_path = ["public", "extensions"]` is preserved, so RPC bodies can still natively reference `public.*` tables — which is exactly the architectural pattern this PR enshrines (read-back inside the RPC body, not over the wire).

Side-effect surface analyzed in commit message and corroborated by my own audit: no realistic regression vector.

### Concern #2 specifics — is the tasks.md discipline doc usable?

Yes — formatted as a checklist line (`- [x] Deploy to dev + smoke against real Supabase + paste log evidence into this card BEFORE opening PR.`) rather than buried in prose. The "Process discipline" header is a reusable template explicitly meant to be copied into future Edge Function / `_shared/` cards. The card's own status of "first formalized DURING this card's hotfix cycle" is a load-bearing narrative anchor — it identifies the cautionary lesson and binds the rule to a concrete PR-#60/#61 trail.

### Things considered but decided not to raise

- Test 8 documents a fail-open posture rather than tightening to fail-closed. I considered raising this as a residual finding, but the "projection is correct; ban is recoverable; product can later flip to fail-closed and Test 8 becomes the test point" framing is sound — and the architecturally important property (the warn log exists; the partial-failure path is now exercised in UAT rather than ignored) is satisfied. Raising fail-closed-vs-fail-open as a new architectural finding would be re-litigation, not verification.
- The "Phase 5 — Verification" checklist in tasks.md still shows unchecked boxes for typecheck/lint/build/etc. The commit message and PR body confirm they all passed; the boxes appear to be the original retrofit phase that predates the SQL-RPC pivot ledger in Phase 7. Not a blocker — Phase 7's `[x] Verify:` line at `:73` is the post-pivot record.

### Bottom line

All seven inline findings and both addressable architecture concerns are fixed correctly. No regressions introduced by the new commit. UAT (especially Test 1, Test 2, and the new Test 8) remains the merge gate per the PR description.

PR #64, opened 2026-05-13
feat(invitation): reject direct cross-provider invitations at the EF boundary

lars-tice:
## Summary

Closes the architectural drift identified in PR #63 UAT Test 5: `accept-invitation` silently allowed provider→provider Sally invitations, emitting `user.role.assigned` events that created native multi-tenant role rows.

Per [`provider-partners-architecture.md`](documentation/architecture/data/provider-partners-architecture.md), cross-tenant access between `type='provider'` orgs is reserved for users whose home org is `type='provider_partner'`, mediated by `cross_tenant_access_grants_projection` — **not** native role assignment.

- New SQL RPC `api.check_invitation_acceptance_eligibility` (read-shape; GRANT to `authenticated` + `service_role`)
- Boundary gates in `accept-invitation` (HTTP 403, Sally path) and `invite-user` (HTTP 422, pre-issuance)
- One-off cleanup: revokes dakaratekid's accidental `Cypress Admin @ testorg-20260329` role (the known victim) inline in the migration

Card: [`dev/active/reject-cross-provider-invitations/`](dev/active/reject-cross-provider-invitations/) (architect-reviewed inline — 5 must-fix items C1–C5 addressed before code).

## Architect review — must-fix items addressed

| # | Item | Resolution |
|---|---|---|
| **C1** | `auth.uid() IS NULL` guard would break `accept-invitation`'s service-role admin client (no JWT subject) | No runtime guard; GRANT to both `authenticated` AND `service_role`, mirroring `api.check_user_invitation_existence` precedent |
| **C2** | Pre-issuance gate target was wrong | Verified via grep that `user.invited` is emitted only at `invite-user/index.ts:836`; gate placed there (NOT `manage-user`) |
| **C3** | Doc enum mismatch (legacy `type='partner'` + `'agency_assignment'`/`'family_consent'`) | Live CHECK is `('platform_owner', 'provider', 'provider_partner')`; DB enum is `'social_services_assignment'`/`'family_participation'`. Doc edits applied |
| **C4** | Eligibility filter blind spots | (a) `organization_id IS NOT NULL` excludes super_admin; (b) `role_valid_from <= today` excludes future-dated; (c) `op.is_active = true` excludes deactivated-org stale roles |
| **C5** | Cleanup idempotency invariant implicit | Documented inline: depends on `handle_user_role_revoked` deleting the projection row |

## Pre-deploy audit (NOTICE-only, run by the migration)

[...]

PR #66, opened 2026-05-19
fix(api): list_users gates membership by accessible_organizations not role-EXISTS

lars-tice:
## Summary

- Fixes `api.list_users` so admins can see "zombie" users — users who joined the org but have no current role rows. Discovered during PR #64 UAT T2 prep (2026-05-18) when `lars.tice+test3@gmail.com` was invisible despite being a known testorg member.
- Adds the GIN index that backs the new `= ANY(accessible_organizations)` predicate.
- Corrects a pre-existing documentation lie about that index name.
- Seeds a follow-up card for the architectural sibling issue (three sister `list_users_*` RPCs gate by `current_organization_id` instead).

## Investigation findings (worth reading)

**Audit of all `list_users*` RPCs:**

| Function | Membership predicate | Has role-EXISTS bug? |
|---|---|---|
| `api.list_users` | EXISTS on user_roles_projection | **YES** (fixed here) |
| `api.list_users_for_role_management` | `u.current_organization_id = v_org_id` | No (different smell — separate card) |
| `api.list_users_for_bulk_assignment` | `u.current_organization_id = v_org_id` | No (different smell — separate card) |
| `api.list_users_for_schedule_management` | `u.current_organization_id = v_org_id` | No (different smell — separate card) |

Per the alignment decision before implementation: literal scope is "fix what shares this exact pattern" → only `api.list_users`. Sisters share a *different* problematic pattern and become a separate card.

**Zombie lineage** of test subject `lars.tice+test3@gmail.com`:
- 2026-05-06: `user.invited` event with `roles: []` (invitation created with no roles)
- 2026-05-11: `user.created` + `invitation.accepted` (lineage `invited_via: organization_bootstrap`)
- Never received a `user.role.assigned` event → permanent zombie state

The "zombie" class can also arise from "role revoked but user not deactivated" — both produce the same invisibility symptom. The visibility fix in this PR is necessary regardless of which lineage created the zombie.

**Documentation drift discovered:** `tables/users.md:185` documented an index `idx_users_roles ON users USING GIN (accessible_organizations)`. Reality: `idx_users_roles` is GIN on `roles` (different column), and there was NO GIN index on `accessible_organizations`. The doc was promising an index that never shipped. This PR ships the promised index (under its correct name `idx_users_accessible_orgs_gin`) and corrects the doc.

## Changes

- **Migration** `20260519233323_fix_list_users_include_roleless.sql`:
  - `CREATE INDEX IF NOT EXISTS idx_users_accessible_orgs_gin ON public.users USING GIN (accessible_organizations);`
  - `CREATE OR REPLACE FUNCTION api.list_users(...)` — replaces both role-EXISTS predicates (count + SELECT) with `p_org_id = ANY(u.accessible_organizations)`. Per-user roles aggregation unchanged.
  - Re-emits `@a4c-rpc-shape: read` COMMENT (defensive against future DROP+CREATE).
- **Docs** `tables/users.md`:
  - Replaces the inaccurate `idx_users_roles` section with the correct `idx_users_accessible_orgs_gin` section.
  - Bumps `last_updated` to 2026-05-19.
- **Follow-up card** `dev/active/list-users-sister-functions-membership-gating/`:
  - Plan + tasks documenting the sister-RPCs `current_organization_id` smell as a separate body of work.

No frontend code change required — `UserCard.tsx:290-311` already renders an amber "No roles assigned" badge for `user.roles.length === 0`.

## Verification

Already done on dev:

- [x] **Smoke**: `api.list_users(testorg_uuid)` returns 3 users (was 2 pre-fix); lars.tice+test3 appears with `roles: []`
- [x] **Regression — `p_status='deleted'`**: returns 2 soft-deleted users (unchanged behavior)
- [x] **Regression — search_term**: `'test3'` returns just lars.tice+test3
- [x] **Regression — pagination**: `p_page_size=2` returns 2 rows with `total_count=3`
- [x] **`@a4c-rpc-shape: read` comment** confirmed via pg_description query
- [x] **GIN index** present on dev (`idx_users_accessible_orgs_gin` via `pg_indexes`)
- [x] Frontend `npm run typecheck && npm run lint && npm run docs:check && npm run build` — all green

Post-merge UI walkthrough (will run once `supabase-migrations.yml` deploys to dev):

- [ ] Log in as a testorg admin
- [ ] Navigate to `/users`
- [ ] Confirm `lars.tice+test3@gmail.com` row appears with amber "No roles assigned" badge
- [ ] Confirm pagination, search, and status filters still work
- [ ] Sanity-check at least one other org's `/users` page to ensure no over-inclusion regression

## Out of scope (per pre-implementation alignment)

- **Sister functions** (`list_users_for_*`) — separate follow-up card seeded
- **`trg_sync_accessible_orgs` trigger refactor** into event handlers — decided "needlessly heavy" during planning (current pattern is strictly safer for replay; would increase drift risk per handler)
- **`current_organization_id` semantics reframing** — large architectural question, parked
- **Soft-deleted user toggle** — out of scope; existing `WHERE u.deleted_at IS NULL` behavior preserved

## Card

`dev/active/users-list-omits-roleless-members/` — will be archived on merge.


infrastructure/supabase/supabase/migrations/20260519233323_fix_list_users_include_roleless.sql
```diff
@@ -0,0 +1,172 @@
...
+-- by u.current_organization_id instead -- a DIFFERENT membership-gating
+-- smell that makes multi-org users invisible in role/schedule admin UIs.
+-- Seeded as a separate follow-up card; out of scope here.
+--
+-- Idempotent: CREATE INDEX IF NOT EXISTS + CREATE OR REPLACE FUNCTION.
+-- ============================================================================
+
+-- GIN index supporting `p_org_id = ANY(u.accessible_organizations)` lookups.
+CREATE INDEX IF NOT EXISTS idx_users_accessible_orgs_gin
+  ON public.users USING GIN (accessible_organizations);
```
lars-tice, line 103: **Must-fix — GIN index will not be used by this predicate shape.**

PostgreSQL's GIN array opclass (`array_ops`) supports the containment operators `@>`, `<@`, `&&`, and `=`. It does **not** support `scalar = ANY(column)` directly — the planner has no rewrite from scalar-equality-via-`ANY` into a GIN-eligible array-containment predicate. As written, both `WHERE p_org_id = ANY(u.accessible_organizations)` sites (L93 and L132) will fall back to a Seq Scan despite this index existing.

The repo's own idiom elsewhere uses the indexable form (e.g., `tags @> ARRAY['development']` in baseline_v4 organizations_projection comment).

**Recommended fix** (single-keystroke change in two places):

```sql
WHERE u.accessible_organizations @> ARRAY[p_org_id]::uuid[]
```

`@>` is the GIN-friendly containment operator; `ARRAY[p_org_id]::uuid[]` keeps the array element type aligned with the column. Verify with `EXPLAIN (ANALYZE, BUFFERS) SELECT * FROM api.list_users(...)` before/after — pre-fix should show `Bitmap Index Scan on idx_users_accessible_orgs_gin`.

If you keep `= ANY(...)` as-is, the index is dead weight (write amplification on every users-table mutation for no read benefit) and the migration comment "Add a GIN index so the common admin-list refresh is an Index Scan rather than a Seq Scan" is false. Either rewrite the predicate or drop the index from this PR.

infrastructure/supabase/supabase/migrations/20260519233323_fix_list_users_include_roleless.sql
```diff
@@ -0,0 +1,172 @@
...
+SET search_path TO 'public', 'api'
+AS $function$
+DECLARE
+  v_total_count BIGINT;
+BEGIN
+  -- Total count for pagination
+  SELECT COUNT(DISTINCT u.id)
+  INTO v_total_count
+  FROM public.users u
+  WHERE p_org_id = ANY(u.accessible_organizations)
```
lars-tice: **Should-consider — visibility predicate has no authorization gate.**

This RPC is `SECURITY DEFINER` (owner `postgres`, GRANT EXECUTE to `authenticated` per baseline L16354) and now exposes every user whose `accessible_organizations` contains `p_org_id` to any caller who can guess a UUID — including PII (`email`, `first_name`, `last_name`, `last_login`). The pre-PR version had the same gap, but the predicate change broadens what's enumerable: every membership now surfaces a row, including roleless users you may not want surfaced cross-tenant.

The original `EXISTS` predicate was an implicit (and unreliable) authorization proxy — you had to have a role in the org to be visible. Now membership alone suffices.

Per `infrastructure/supabase/CLAUDE.md` § "Choosing between `has_permission()` and `has_effective_permission()`": user-as-identity has no organizational location finer than tenant, so the tenancy guard pattern applies — check that `p_org_id = get_current_org_id()` OR `has_platform_privilege()` at the top of the function, returning empty otherwise. Pattern is used by `api.deactivate_user`, `api.delete_user`, etc.

```sql
IF NOT (
  has_platform_privilege()
  OR p_org_id = get_current_org_id()
) THEN
  RETURN;  -- empty result set, indistinguishable from "no members"
END IF;
```

Verdict-bumping if undefended: this is a SECURITY DEFINER RPC granted to `authenticated` that enumerates PII keyed only by a caller-supplied org UUID.

infrastructure/supabase/supabase/migrations/20260519233323_fix_list_users_include_roleless.sql
```diff
@@ -0,0 +1,172 @@
...
+      ))
+      FROM public.user_roles_projection ur
+      JOIN public.roles_projection r ON r.id = ur.role_id
+      WHERE ur.user_id = u.id
+        AND ur.organization_id = p_org_id),
+      '[]'::jsonb
+    ) AS roles,
+    v_total_count AS total_count
+  FROM public.users u
+  WHERE p_org_id = ANY(u.accessible_organizations)
```
lars-tice: **Should-consider — predicate duplication is a drift hazard.**

The same complex predicate block (membership + status CASE + search ILIKE) is duplicated between L90–104 (count) and L131–143 (SELECT). If the next maintainer changes one (e.g., to add a `p_include_invitations` filter) and misses the other, `total_count` will lie and pagination math will silently break.

Two safer shapes that the PR could adopt with minimal scope creep:
1. **Window function**: `COUNT(*) OVER ()` in the SELECT replaces the separate count subquery entirely (single source of truth for the predicate).
2. **CTE**: `WITH filtered_users AS (...)` materializes once and the count/page draw from the same row source.

Not a blocker — the duplication exists in the prior baseline body too — but flagging because the PR is the natural place to fix it now that you're rewriting the predicate anyway. If you'd rather defer, leave a comment block calling out the duplication invariant ("if you change this WHERE, also change L90–104").

infrastructure/supabase/supabase/migrations/20260519233323_fix_list_users_include_roleless.sql
```diff
@@ -0,0 +1,172 @@
...
+END;
+$function$;
+
+-- Per infrastructure-guidelines Rule 17 / supabase/CLAUDE.md "RPC Shape Registry":
+-- re-emit the RPC shape tag. CREATE OR REPLACE preserves the existing COMMENT,
+-- but defensive re-issue guards against future DROP+CREATE losing it.
+COMMENT ON FUNCTION api.list_users(uuid, text, text, text, boolean, integer, integer) IS
+$comment$List users in an organization with pagination and filtering. Membership gated by users.accessible_organizations (the canonical membership oracle, maintained by trg_sync_accessible_orgs from user_organizations_projection). Status values: active, deactivated, deleted (or NULL for all non-deleted).
+
+@a4c-rpc-shape: read$comment$;
```
lars-tice, line 210: **Nit (Rule 17 compliance — defensive re-emission verified).**

Confirmed: `@a4c-rpc-shape: read` tag matches the prior `apiRpc<RpcUserRow[]>('list_users', ...)` call site at `frontend/src/services/users/SupabaseUserQueryService.ts:316`. Signature is unchanged so `CREATE OR REPLACE` preserves the OID and the comment would have survived without re-emission — but the defensive re-issue is exactly the discipline Rule 17 asks for. Good.

One picky observation: the prior baseline comment did NOT carry the `@a4c-rpc-shape` tag at all (baseline L4683 lacks it). That's an undocumented pre-existing inventory gap — `pg_description.description` was already correct on dev because some later migration must have added it, OR `UncategorizedRpcs` would have caught `list_users` in CI. Worth a one-line check post-merge that `frontend/src/services/api/rpc-registry.generated.ts` still lists `list_users` under `ReadRpcs` (it does as of HEAD, L173). No action required if the CI gate passes.

frontend/docs-validation-report.json
```diff
@@ -129,7 +129,7 @@
     "componentsDocumented": 0,
     "apisDocumented": 0
   },
-  "timestamp": "2026-05-11T19:32:42.708Z",
+  "timestamp": "2026-05-19T23:39:02.499Z",
```
lars-tice: **Nit — generated artifact in the diff.**

This file is the output of `npm run docs:check` and the only change is a timestamp bump (`2026-05-11T19:32:42.708Z` → `2026-05-19T23:39:02.499Z`). Looking at git history, this file IS canonically committed (prior commits do touch it), so this is not a leak — but it's still a noisy churn entry that obscures the load-bearing diff.

Two options for the future, neither blocking for this PR:
1. Add `frontend/docs-validation-report.json` to `.gitignore` and stop tracking it (treat it as a build artifact; CI can regenerate and assert).
2. Modify the generator to omit the `timestamp` field (or pin to a stable value) when the report content is unchanged — so commits only churn the file when something meaningful drifted.

Leaving as-is is fine; flagging only because the PR description doesn't mention this file and a reviewer might wonder why a "frontend" artifact lands in a backend-only PR.

documentation/infrastructure/reference/database/tables/users.md
```diff
@@ -180,16 +180,18 @@ CREATE INDEX idx_users_current_organization ON users(current_organization_id);
...
+CREATE INDEX idx_users_accessible_orgs_gin ON public.users USING GIN (accessible_organizations);
 ```
 - **Purpose**: Enable efficient queries on accessible organizations array
 - **Type**: GIN (Generalized Inverted Index) for array containment
 - **Usage**:
   - Find all users with access to an organization: `WHERE '<org-uuid>' = ANY(accessible_organizations)`
+  - Membership oracle queries in `api.list_users` and any other admin-list RPC
   - Multi-tenant access queries
-- **Performance**: Essential for array containment queries
+- **Performance**: Essential for array containment queries — without this index, predicates of the form `<uuid> = ANY(accessible_organizations)` degrade to a sequential scan
```
lars-tice: **Nit — doc-drift correction is accurate, with a follow-on suggestion.**

Spot-checked the doc-drift claim. Confirmed: baseline_v4 L14823 has `CREATE INDEX "idx_users_roles" ON "public"."users" USING "gin" ("roles")` — the existing `idx_users_roles` index is GIN on the `roles` JSONB column, **not** on `accessible_organizations`. The pre-PR doc text was a lie about an index that never shipped. PR correctly replaces it with the truthful `idx_users_accessible_orgs_gin`. `last_updated` bumped to 2026-05-19. Good.

Follow-on for a future doc pass (out of scope for this PR): the indexes section of `users.md` is missing the real `idx_users_roles` (GIN on `roles`) entirely. Now that this PR is touching the section, consider adding the real one in the same edit, or seed a one-line doc card. Otherwise the table doc continues to underrepresent the actual user-table index set.

dev/active/list-users-sister-functions-membership-gating/plan.md
```diff
@@ -0,0 +1,64 @@
...
+| `api.list_users` | `p_org_id = ANY(u.accessible_organizations)` | ✅ Correct (fixed in PR #66) |
+| `api.list_users_for_role_management` | `u.current_organization_id = v_org_id` | ⚠️ Smell — this card |
+| `api.list_users_for_bulk_assignment` | `u.current_organization_id = v_org_id` | ⚠️ Smell — this card |
+| `api.list_users_for_schedule_management` | `u.current_organization_id = v_org_id` | ⚠️ Smell — this card |
+
+## Problem statement
+
+The three sister functions use `u.current_organization_id = v_org_id` as the membership gate. `current_organization_id` is the user's *currently selected* org (their "active session" org), NOT a membership oracle. This means:
+
+1. **Multi-org users**: a user with home org `liveforlife` who also has a role assignment in `testorg` is invisible to testorg's role-management UI because their `current_organization_id` is liveforlife.
```
lars-tice, line 18: **Should-consider — sister-card scope decision is defensible but rebuttable.**

The card-vs-inline call here hits the bar set 2026-05-06 in memory ("default to in-PR fixes; cards are for genuinely separate bodies of work"). Arguments for KEEPING the sister card:

- The bug class is different: `current_organization_id = v_org_id` is a *session-pointer-as-membership* smell, not a *role-EXISTS-as-membership* smell. The sister fix has its own UAT design (multi-org test subject construction).
- The sister functions accept `p_scope_path ltree`/`p_role_id`/`p_template_id` — different surface area, different review concerns (RBAC scope path semantics that this PR doesn't touch).
- A unified-visibility PR would land in three places at once, with different UI walkthroughs — more risk per merge.

Argument for IN-PR rollup: the unified-visibility model (everywhere uses `accessible_organizations`) is one architectural shift; shipping it piecemeal risks the same drift the PR is repairing in `list_users`.

On balance the card seed is fine — but the plan.md should call out explicitly that the GIN-friendliness of the predicate shape (see migration L58 comment) carries through to the sister PR, and reference whatever resolution this PR lands on (`@>` vs `= ANY`). Otherwise the sister PR will repeat the same predicate-shape decision in isolation.

lars-tice (commented):
# Architectural Review — PR #66

**Verdict: REQUEST-CHANGES (one Must-fix; one Should-consider; rest are non-blocking)**

> _Note: posted as `--comment` rather than `--request-changes` because the reviewer is the PR author — GitHub blocks self-request-changes. The verdict-bearing content is unchanged: the GIN-index issue is load-bearing and the predicate shape needs to change OR the index needs to come out before merge._

The semantic fix is correct: replacing role-EXISTS with `accessible_organizations` membership is the right architectural move, the membership oracle (`trg_sync_accessible_orgs`) is the load-bearing reason the predicate substitution is safe, and the doc-drift correction in `tables/users.md` is accurate. But the GIN index does not back the chosen predicate shape, and the new SECURITY DEFINER predicate exposes more PII than the prior one.

## Findings

### Must-fix

1. **GIN index will not be used by `scalar = ANY(column)`.** PostgreSQL's GIN `array_ops` opclass indexes `@>`, `<@`, `&&`, `=` — not the `ANY` form. The repo's own idiom uses `@>` (baseline_v4 organizations_projection comment). Either rewrite predicate to `u.accessible_organizations @> ARRAY[p_org_id]::uuid[]` (two sites: L93, L132) **or** drop the index from this PR. As-is: index is dead weight + the migration's `Performance` rationale is false. Verify with EXPLAIN before merge. (Inline @ L58.)

### Should-consider

2. **SECURITY DEFINER RPC has no tenancy gate.** Pre-PR, the role-EXISTS predicate was an implicit (and unreliable) authorization proxy. Now `p_org_id` membership alone surfaces PII (email/name/last_login) to `authenticated`. Per `infrastructure/supabase/CLAUDE.md` § scoped-vs-unscoped, add `IF NOT (has_platform_privilege() OR p_org_id = get_current_org_id()) THEN RETURN; END IF;` at the top — matches the `api.deactivate_user` / `api.delete_user` tenancy-guard pattern. (Inline @ L93.)

3. **Predicate duplication invariant.** The membership + status + search filter block is duplicated between L90–104 (count) and L131–143 (SELECT). Drift risk on every future filter change. Either rewrite as `COUNT(*) OVER ()` window or a CTE; or add a comment block calling out the duplication invariant. (Inline @ L132.)

4. **Sister-card scope decision is defensible.** Acceptable per the "in-PR fixes vs cards" rule given the different bug class (session-pointer vs role-EXISTS) and the surface difference (ltree scope path semantics). Plan.md should reference whatever predicate-shape resolution this PR lands on (`@>` vs `= ANY`) so the sister fix doesn't re-litigate. (Inline @ plan.md L18.)

### Nits

5. **Rule 17 (`@a4c-rpc-shape: read`) — verified.** Defensive re-issue is correct. Worth a one-line post-merge sanity check that `rpc-registry.generated.ts` still resolves `list_users` to `ReadRpcs`. (Inline @ L172.)

6. **Doc-drift fix verified.** `idx_users_roles` in baseline is GIN on `roles` (not `accessible_organizations`); PR corrects to truth. Follow-on: indexes section still lacks the real `idx_users_roles` entry. (Inline @ users.md L193.)

7. **`frontend/docs-validation-report.json`** — committed canonically per history; only churning the timestamp. Future ergonomics: gitignore or pin timestamp. (Inline @ that file.)

## Verification of PR claims

- ✅ UserCard.tsx:290–311 renders amber "No roles assigned" badge for empty `user.roles` — verified at L308.
- ✅ `accessible_organizations` is maintained by `trg_sync_accessible_orgs` from `user_organizations_projection` — verified in baseline_v4 L11767–11790, L14864.
- ✅ Predicate substitution is membership-safe: the sync trigger recomputes the array from current projection state, so a role-row alone does not add to `accessible_organizations` and a removed-membership cannot leave stale array entries.
- ✅ Re-emitted `COMMENT ON FUNCTION ... @a4c-rpc-shape: read` survives `CREATE OR REPLACE` (same OID); defensive re-issue is correct per Rule 17.
- ⚠️ `handle_user_invited` (baseline L9210–9212) writes `accessible_organizations` via array union BEFORE `user_organizations_projection` is INSERTed (L9216) — both rows are written in the same handler so the sync trigger reconciles immediately, but a future refactor that splits these into separate event handlers could produce a window where the array differs from the projection. Not a blocker for this PR; worth a comment in `sync_accessible_organizations` reference doc.

## Architecture Review Checklist

- [x] **CQRS Standards** — PASS. RPC is the wire entry point; no direct table embedding.
- [x] **Naming Conventions** — PASS. Index name follows `idx_<table>_<col>_<type>` convention.
- [PARTIAL] **Design Patterns** — Read-side projection query, correct pattern. Tenancy-guard pattern is missing (see Finding 2).
- [N/A] **data-testid** — no UI change.
- [N/A] **AsyncAPI Event Registration** — no new events.
- [x] **Type Generation** — PASS. Signature unchanged; `RpcUserRow[]` type at frontend matches RETURNS TABLE.
- [PARTIAL] **Observability / Tracing** — pre-existing gap; this RPC has no `request_id` propagation. Out of scope here.
- [N/A] **Error Surfacing to UI** — no error path in a read RPC.
- [FAIL] **Performance / Indexing** — index in PR will not be used by the predicate shape (Finding 1).
- [x] **Rule 17** — PASS, defensive re-issue.
- [N/A] **Rule 19** — no Edge Function cross-schema reads.

## Punch list (merge blockers)

- [ ] **Finding 1 (Must-fix)**: Rewrite predicate to `accessible_organizations @> ARRAY[p_org_id]::uuid[]` at both sites, or drop the GIN index. Verify with EXPLAIN.
- [ ] **Finding 2 (Should-fix in-PR per the 2026-05-06 rule)**: Add `has_platform_privilege() OR p_org_id = get_current_org_id()` tenancy guard.
- [ ] Re-run dev smoke (3 users; regressions) after the above two changes.

Findings 3, 4, 5, 6, 7 are non-blocking — fold in opportunistically or defer.

lars-tice:
## Review-response summary — all 7 findings addressed

Pushed `21886472` to address every finding from the 2026-05-20 architectural review. Per-finding resolution:

### Must-fix

**Finding 1 — GIN index unused by `= ANY`** ✅
- Rewrote both predicate sites to `u.accessible_organizations @> ARRAY[p_org_id]::uuid[]`
- EXPLAIN verification (with `SET enable_seqscan=off` to force the planner; dev table is too small to trigger index naturally):
  ```
  Bitmap Heap Scan on users
    Recheck Cond: (accessible_organizations @> '{2d0829ae-...}'::uuid[])
    ->  Bitmap Index Scan on idx_users_accessible_orgs_gin
          Index Cond: (accessible_organizations @> '{2d0829ae-...}'::uuid[])
  ```

### Should-consider

**Finding 2 — Tenancy guard** ✅
- Added at top of function body:
  ```sql
  IF NOT (
    public.has_platform_privilege()
    OR p_org_id = public.get_current_org_id()
  ) THEN
    RETURN;
  END IF;
  ```
- Verified via JWT-claim simulation: caller in liveforlife querying for testorg → 0 rows; caller in testorg querying for testorg → 3 rows.

**Finding 3 — Predicate duplication** ✅
- Refactored to `COUNT(*) OVER ()` window function. The membership + status + search predicate now lives in exactly one place.
- Precedent followed: `api.get_organizations_paginated()` (baseline L3181 + `20260306214844` L160).
- Removed the `v_total_count BIGINT;` declaration and the entire count subquery block.

**Finding 4 — Sister-card plan reference** ✅
- `dev/active/list-users-sister-functions-membership-gating/plan.md` "Fix sketch" section now cites the `@>` form as the convention established by this PR, with a brief rationale (GIN `array_ops` opclass).

### Nits

**Finding 5 — Post-merge rpc-registry verification** ✅ (no code change)
- The `rpc-registry-sync.yml` CI workflow already gates that `list_users` remains in `ReadRpcs`. Green check on merge is the sufficient evidence.

**Finding 6 — Missing `idx_users_roles` doc entry** ✅
- Added a new `#### idx_users_roles` section to `tables/users.md` documenting the real GIN-on-roles JSONB index.
- Also added a predicate-shape warning to the `idx_users_accessible_orgs_gin` section so future contributors don't repeat the `= ANY` trap.
- Bumped `last_updated: 2026-05-20`.

**Finding 7 — `docs-validation-report.json` timestamp churn** ✅
- `git rm --cached frontend/docs-validation-report.json` + added `docs-validation-report.json` to `frontend/.gitignore`.
- Per memory rule 18 (gitignore is per-directory), the pattern is in the frontend gitignore specifically.

### Soft note from verification section ✅
The reviewer's "verification of PR claims" flagged the `handle_user_invited` two-write ordering. No reference file exists for `sync_accessible_organizations` (only event handlers have reference files, not trigger functions). Addressed by adding a 3-line acknowledgement block in the migration header so a future contributor touching that path will see the invariant flagged.

### Deploy + verification
- `supabase migration repair --status reverted 20260519233323` + `supabase db push --linked` (per memory `pr-63-close-out.md` — safe on feature branches not yet merged)
- All 6 verification checks pass on dev (functional smoke, EXPLAIN with index usage, tenancy guard both directions, status filter, search filter)
- Frontend gate: `typecheck && lint && docs:check && build` all green

### Out of scope (acknowledged, not addressed here)
- **Reference file for `sync_accessible_organizations`** — no `handlers/` precedent for trigger functions; would be a new convention. Acknowledged via migration header comment instead.
- **Generator-side fix for `docs-validation-report.json` timestamp field** (option 2) — bigger blast radius; untracking (option 1, chosen) is cheaper and ends the timestamp-churn class entirely.
- **Sister `list_users_for_*` functions** — separate card seeded at `dev/active/list-users-sister-functions-membership-gating/` (plan now cites the `@>` convention established here).

Ready for re-review.

lars-tice (commented):
# Second architect review — fix commit `21886472`

**Verdict**: **APPROVE WITH ONE OPTIONAL IN-PR FIX** (non-blocking)

The author's second commit substantively addresses all 7 findings from the first review. Verified by reading each file at head SHA `21886472` against the migration body, the docs change, the gitignore, and the helper definitions in `baseline_v4.sql`. The migration is cleaner than the first commit on every axis — predicate, tenancy, and code-duplication.

---

## Per-finding verification

| # | Severity | Status | Evidence |
|---|----------|--------|----------|
| 1 | Must-fix | ✅ Resolved | Migration L170: `WHERE u.accessible_organizations @> ARRAY[p_org_id]::uuid[]`. **Single occurrence** (F3 collapsed the duplicate). No `= ANY(u.accessible_organizations)` remains in the function body. GIN index `idx_users_accessible_orgs_gin` still present at L102-103 (`IF NOT EXISTS`, idempotent). |
| 2 | Should-consider → in-PR | ✅ Resolved | Migration L135-140: guard is the first executable statement after `BEGIN`, before any query work. Returns empty result set (matches `RETURNS TABLE` contract, no envelope drift). Helper names verified: `public.has_platform_privilege()` (baseline L9957) and `public.get_current_org_id()` (baseline L7436). **OR-ordering is correct for super_admin/NULL-org-id case**: `has_platform_privilege()` reads `effective_permissions` from JWT claims (doesn't require `org_id`), so a super_admin querying any tenant subdomain — even with `org_id=NULL` in JWT (the symptom captured by the `superadmin-no-org-context-on-tenant-subdomain` card) — short-circuits TRUE before `p_org_id = NULL` is ever evaluated. |
| 3 | Should-consider | ✅ Resolved | Single SELECT, single WHERE clause. `COUNT(*) OVER ()` at L168. Window-function semantics correct: count is over the filtered set BEFORE `LIMIT/OFFSET`, so pagination math is sound. Every returned row carries the same `total_count` value (consumer can read row[0].total_count or aggregate — same answer). Cited precedent verified: `baseline_v4.sql:3167-3181` (`api.get_organizations_paginated`) uses the identical `COUNT(*) OVER()` pattern, including the empty-rowset-implies-zero-count contract. |
| 4 | Should-consider | ✅ Resolved | `dev/active/list-users-sister-functions-membership-gating/plan.md:42-45` now reads `WHERE u.accessible_organizations @> ARRAY[v_org_id]::uuid[]` and explicitly cross-links to "PR #66 architectural review, Finding 1" for the predicate-shape rationale. |
| 5 | Nit | ✅ Resolved | `COMMENT ON FUNCTION ... @a4c-rpc-shape: read` re-emitted at migration L207-210. CI workflow `.github/workflows/rpc-registry-sync.yml` exists and gates on `UncategorizedRpcs` non-empty + registry drift; would catch any future regression. |
| 6 | Nit | ✅ Resolved (with one residual) | `documentation/infrastructure/reference/database/tables/users.md` frontmatter `last_updated: 2026-05-20` (L3). `idx_users_accessible_orgs_gin` section at L183-195 includes the predicate-shape warning explicitly calling out `= ANY` as GIN-incompatible. `idx_users_roles` section at L197-205 documents the legacy JSONB-on-roles index and explicitly notes the naming-confusion vs the membership index. **See residual below.** |
| 7 | Nit | ✅ Resolved | `git ls-tree 21886472 -- frontend/docs-validation-report.json` returns empty (untracked). `frontend/.gitignore:99` contains `docs-validation-report.json`. |

---

## Residual (non-blocking) — Nit-6b

`documentation/infrastructure/reference/database/tables/users.md:715-729` ("Troubleshooting → Array Containment Query Slow") is a pre-existing section that **wasn't reconciled** with the F6 surface-level addition:

```
**Expected Plan**: Should use `idx_users_roles` GIN index    ← wrong index name
...
- REINDEX if needed: `REINDEX INDEX idx_users_roles;`        ← wrong index name
```

And the EXPLAIN example on L722 uses `= ANY(accessible_organizations)` — the exact pattern F1 fixed and the new section warns against.

**Recommendation**: in-PR follow-up edit — change the index name in the troubleshooting section to `idx_users_accessible_orgs_gin` and rewrite the EXPLAIN example to `WHERE accessible_organizations @> ARRAY['<org-uuid>'::uuid]`. Single-paragraph change. Cheaper to fix now than carve another card.

This is **optional** — the new index section above the troubleshooting block is internally consistent and authoritative, and a reader following the new entry will write correct predicates. The troubleshooting section is a discoverability hazard, not a correctness regression.

---

## New findings introduced by the fix commit

**None.** I checked:
- Window-function over LIMIT/OFFSET semantics (correct — counts pre-pagination)
- Tenancy guard short-circuit for platform admins with NULL org_id (correct — `has_platform_privilege()` doesn't depend on `org_id`)
- The `handle_user_invited` two-write ordering note in the migration header (accurate — `baseline_v4.sql:9210-9216` does perform the array union and the `user_organizations_projection` INSERT within the same handler invocation; the "future refactor" caveat is appropriately framed as forward-guard, not a present-day defect)
- Idempotency (CREATE INDEX IF NOT EXISTS + CREATE OR REPLACE FUNCTION — clean)
- Search path (`SET search_path TO 'public', 'api'` — appropriate for SECURITY DEFINER)

---

## Architecture review checklist

- **CQRS Standards**: ✅ Read RPC, no projection writes; `@a4c-rpc-shape: read` tag correctly applied.
- **Naming Conventions**: ✅ Helper names match codebase; index name follows existing convention (`idx_<table>_<column>_<type>`).
- **Design Patterns**: ✅ Tenancy guard via has-permission OR self-org match is the canonical pattern documented in `infrastructure/supabase/CLAUDE.md` § "Choosing between has_permission and has_effective_permission". `COUNT(*) OVER ()` is the canonical paginated-read pattern (precedent: `api.get_organizations_paginated`).
- **data-testid attributes**: N/A (no UI change).
- **AsyncAPI**: N/A (no new event).
- **Type generation**: ✅ Function signature unchanged; `database.types.ts` regen not required.
- **Observability / tracing**: N/A (no new code path; `RETURN;` for cross-tenant is intentionally indistinguishable from "no members" by design — see migration header).
- **Error surfacing to UI**: ✅ Empty result set is correct UX (admin sees empty list, no leak of "tenant exists but you can't see it" signal).

---

## Punch list

**Blocker**: none.
**Optional in-PR fix** (nit-6b): reconcile `tables/users.md:715-729` troubleshooting section with the new index/predicate convention. Single-paragraph edit.

Merge gate is clear regardless of whether 6b lands in this PR or a doc-cleanup follow-up. My recommendation per the 2026-05-06 no-deferral rule: fold it into this PR — it's three lines and in the same file you already touched.

Excellent turnaround. Approved.

PR #67, opened 2026-05-21
fix(api): list_users sister RPCs gate membership by accessible_organizations not current_organization_id

lars-tice:
## Summary

Three sister RPCs to \`api.list_users\` carried a different but related smell from PR #66's fix: they gated user-list membership by \`u.current_organization_id = v_org_id\`. \`current_organization_id\` is the user's **active session pointer**, not a membership oracle. Multi-org users were invisible in role-management, bulk-assignment, and schedule-assignment admin UIs unless their session pointer happened to match the target org.

This PR applies PR #66's \`accessible_organizations @> ARRAY[v_org_id]::uuid[]\` convention to all three:

- \`api.list_users_for_role_management\`
- \`api.list_users_for_bulk_assignment\`
- \`api.list_users_for_schedule_management\`

Reuses \`idx_users_accessible_orgs_gin\` from PR #66; no new index needed.

## Architect review

Plan was architect-reviewed before implementation — verdict **APPROVE WITH FINDINGS** (1 must-fix + 4 should-considers + 4 nits, all folded into the plan before execution). See \`dev/active/list-users-sister-functions-membership-gating/tasks.md\` "Architect-review punch list" for the resolution table.

Notably: the architect's must-fix (Phase 3 smoke synthesis was leaking the projection-as-read-model invariant) drove the transactional \`BEGIN; ... ROLLBACK;\` harness used during verification — see "Smoke verification" below.

## Smoke verification (transactional, zero externally observable drift)

Synthesized \`lars.tice+test2@gmail.com\` multi-org membership (testorg → also liveforlife) inside a single Management API SQL transaction with JWT-claim simulation (dakaratekid's effective_permissions shape), ran all three RPCs against liveforlife, then ROLLBACK. Post-rollback verification confirms zero drift.

| Verification | Result |
|---|---|
| In-tx INSERT visible; trigger reconciled \`accessible_organizations\` | \`[testorg, liveforlife]\` ✓ |
| Test subject \`current_organization_id\` unchanged | Still \`testorg\` ✓ |
| \`list_users_for_role_management\` surfaces test subject | ✓ (\`is_assigned: false\`, \`current_roles: []\`) |
| \`list_users_for_bulk_assignment\` surfaces test subject | ✓ (\`is_already_assigned: false\`) |
| \`list_users_for_schedule_management\` surfaces test subject | ✓ (\`is_assigned: false\`) |
| Regression: existing liveforlife members (dakaratekid, rachel, troy) | ✓ all still appear with correct flags |
| Post-ROLLBACK projection state | Identical to pre-flight (0 rows for test2+liveforlife in user_organizations_projection) |

## Migration session search_path note

Parameter type \`ltree\` lives in the \`extensions\` schema. The function-body \`SET search_path TO 'public', 'extensions', 'pg_temp'\` clauses apply INSIDE the function body but NOT during \`CREATE OR REPLACE\`'s parameter type resolution. First push failed with \`type ltree does not exist (SQLSTATE 42704)\`. Fix: added \`SET search_path = public, extensions, pg_temp;\` at the top of the migration so \`ltree\` resolves during parsing. This pattern may be worth noting in the supabase-CLAUDE.md for future ltree-parameter migrations.

## Out of scope (per architect-approved plan)

- **Super_admin invisibility**: super_admins have \`current_organization_id IS NULL\` AND \`accessible_organizations IS NULL\` (note: live state on dev is \`NULL\`, not \`[]\` as the PR #66 close-out memory states — that's a memory-file accuracy note for next groom). Both pre-fix and post-fix predicates exclude them. They have global permissions, not narrow-scope \`role_id\` rows, so this is acceptable.
- **\`COUNT(*) OVER ()\` dedup pattern from PR #66**: intentionally NOT applied here. These three RPCs don't return \`total_count\`; pagination total is a consumer-side concern. Preempts a "you should be consistent" review comment.
- **Reframing \`current_organization_id\` semantics**: stays as the active-session pointer for direct-care use.

## Docs touched (small adjacent edits)

- \`documentation/architecture/authorization/rbac-architecture.md\`: fixed pre-existing signature/return-shape drift on the \`list_users_for_role_management\` entry (was showing 1-of-5 params and 4 wrong return columns) while adjacent to the change. Added cross-reference to the sister functions. Bumped \`last_updated\`.
- \`documentation/infrastructure/reference/database/tables/users.md\`: updated \`idx_users_accessible_orgs_gin\` section to note it's now load-bearing for all four \`list_users*\` RPCs (was 1 post-PR-66). Bumped \`last_updated\`.

## Frontend impact

No code change required — signatures unchanged. \`database.types.ts\` and \`rpc-registry.generated.ts\` byte-identical (verified via \`npm run typecheck && npm run lint && npm run docs:check && npm run build\` — all green).

## Card

\`dev/active/list-users-sister-functions-membership-gating/\` — will be archived on merge.

### DIFF
--- a/documentation/architecture/authorization/rbac-architecture.md
+++ b/documentation/architecture/authorization/rbac-architecture.md
@@ -1,6 +1,6 @@
 ---
 status: current
-last_updated: 2026-05-06
+last_updated: 2026-05-21
 ---
 
 <!-- TL;DR-START -->
@@ -491,20 +491,30 @@ For bulk operations where administrators need to add and remove multiple role as
 
 #### `api.list_users_for_role_management`
 
-Returns all users in the organization with their current assignment status for a specific role.
+Returns all users in the organization with their current assignment status for a specific role at a specific scope. Membership gated by `users.accessible_organizations` (the canonical membership oracle, maintained by `trg_sync_accessible_orgs` from `user_organizations_projection`). Permission gated by `has_permission('user.role_assign')` within the requested scope path.
 
 ```sql
 -- Signature
-api.list_users_for_role_management(p_role_id uuid)
+api.list_users_for_role_management(
+  p_role_id     uuid,
+  p_scope_path  ltree,
+  p_search_term text    DEFAULT NULL,
+  p_limit       integer DEFAULT 100,
+  p_offset      integer DEFAULT 0
+)
 RETURNS TABLE (
-  user_id uuid,
-  email text,
-  name text,
-  is_assigned boolean  -- True if user currently has this role
+  id            uuid,
+  email         text,
+  display_name  text,            -- COALESCE(u.name, u.email)
+  is_active     boolean,
+  current_roles text[],          -- All role names this user currently holds
+  is_assigned   boolean          -- True if user currently has THIS role at THIS scope
 )
 ```
 
 **Usage**: Called when opening the role assignment dialog to show all eligible users with their current assignment state.
+
+**Sister functions** (`api.list_users_for_bulk_assignment`, `api.list_users_for_schedule_management`) share the same membership-oracle pattern. See migration `20260521195657_fix_list_users_sister_functions_membership_gating.sql` for the unified-visibility predicate convention established across all four `list_users*` RPCs (origin: PR #66).
 
 #### `api.sync_role_assignments`
 
--- a/documentation/architecture/authorization/rbac-architecture.md
+++ b/documentation/architecture/authorization/rbac-architecture.md
@@ -2,6 +2,8 @@
 status: current
 last_updated: 2026-05-21
 ---
+<!-- last_updated note: also touched on 2026-05-21 for PR #67 review-response — permission-gate description updated to has_effective_permission per architect Finding #5 -->
+
 
 <!-- TL;DR-START -->
 ## TL;DR
@@ -491,7 +493,7 @@ For bulk operations where administrators need to add and remove multiple role as
 
 #### `api.list_users_for_role_management`
 
-Returns all users in the organization with their current assignment status for a specific role at a specific scope. Membership gated by `users.accessible_organizations` (the canonical membership oracle, maintained by `trg_sync_accessible_orgs` from `user_organizations_projection`). Permission gated by `has_permission('user.role_assign')` within the requested scope path.
+Returns all users in the organization with their current assignment status for a specific role at a specific scope. Membership gated by `users.accessible_organizations` (the canonical membership oracle, maintained by `trg_sync_accessible_orgs` from `user_organizations_projection`). Permission gated by `has_effective_permission('user.role_assign', p_scope_path)` (verifies the caller holds the permission at a scope that contains the requested path).
 
 ```sql
 -- Signature
--- a/infrastructure/supabase/supabase/migrations/20260521195657_fix_list_users_sister_functions_membership_gating.sql
+++ b/infrastructure/supabase/supabase/migrations/20260521195657_fix_list_users_sister_functions_membership_gating.sql
@@ -1,0 +1,428 @@
+-- ============================================================================
+-- Migration: fix_list_users_sister_functions_membership_gating
+-- Origin:    dev/active/list-users-sister-functions-membership-gating/
+-- Precedent: PR #66 (merged 2026-05-20, commit 33e77a4f) — established the
+--            `accessible_organizations @> ARRAY[<uuid>]::uuid[]` convention
+--            for membership-gated user-listing RPCs.
+-- ============================================================================
+--
+-- Problem
+-- -------
+-- Three sister RPCs to api.list_users carry a different but related smell:
+-- they gate user-list membership by `u.current_organization_id = v_org_id`.
+-- `current_organization_id` is the user's ACTIVE SESSION pointer (per memory,
+-- set via api.switch_org_unit at clock-in for direct-care staff; NULL for
+-- super_admin and platform-only users). It is NOT the membership oracle.
+--
+-- Consequence pre-fix: multi-org users (users with a row in
+-- user_organizations_projection for multiple orgs) are invisible in
+-- role-management, bulk-assignment, and schedule-assignment admin UIs unless
+-- their `current_organization_id` happens to match the target org. Same gap
+-- as PR #66's api.list_users, on three additional surfaces.
+--
+-- Today on dev the defect is dormant (no multi-org users exist), but it
+-- will become routine when the cross-tenant grant pipeline ships per
+-- dev/active/sub-tenant-admin-design/.
+--
+-- Fix
+-- ---
+-- Replace the predicate in all three function bodies:
+--
+--   BEFORE: WHERE u.current_organization_id = v_org_id
+--   AFTER:  WHERE u.accessible_organizations @> ARRAY[v_org_id]::uuid[]
+--
+-- `users.accessible_organizations` (uuid[]) is the canonical membership
+-- oracle, maintained by trigger trg_sync_accessible_orgs from
+-- user_organizations_projection. Same predicate shape PR #66 established
+-- for api.list_users; reuses the GIN index idx_users_accessible_orgs_gin
+-- created in that PR.
+--
+-- Predicate-shape note (re-stated from PR #66): PostgreSQL's GIN array_ops
+-- opclass indexes the containment operators (@>, <@, &&, =) but NOT
+-- `scalar = ANY(col)`. The `@>` form is the GIN-indexable shape; do not
+-- rewrite to `= ANY` without dropping the index.
+--
+-- Scope decisions (per architect-reviewed plan)
+-- ---------------------------------------------
+-- - NOT changing the per-function permission gates (each RPC already has
+--   one: has_effective_permission('user.role_assign', p_scope_path) for the
+--   two role-functions; has_effective_permission('user.schedule_manage',
+--   <ou_or_org_path>) for schedule_management). The predicate swap broadens
+--   what's enumerable AT a given scope, but the scope itself is still
+--   gated by the existing checks.
+-- - NOT reframing `current_organization_id` semantics — it remains the
+--   active-session pointer for direct-care use.
+-- - NOT addressing super_admin invisibility. Super_admins on dev have
+--   `current_organization_id IS NULL` AND `accessible_organizations IS NULL`.
+--   `NULL::uuid[] @> ARRAY[<uuid>]::uuid[]` returns NULL (treated as
+--   not-matching by WHERE), so super_admins remain excluded post-fix.
+--   They have global permissions and rarely have specific role_id
+--   assignments at narrow ltree paths, so this is acceptable. If UX
+--   feedback later disagrees, file a separate card.
+-- - NOT applying PR #66's COUNT(*) OVER () dedup pattern: these three RPCs
+--   don't return total_count; they paginate directly with LIMIT/OFFSET.
+--   Pagination total is a consumer-side concern.
+--
+-- COMMENT prose
+-- -------------
+-- - list_users_for_bulk_assignment had descriptive COMMENT prose pre-fix:
+--   edit to mention `accessible_organizations` as the membership oracle.
+-- - list_users_for_role_management and list_users_for_schedule_management
+--   carried only the bare `@a4c-rpc-shape: read` tag pre-fix: author full
+--   new prose blocks (purpose + membership oracle + permission gate +
+--   shape tag), modeled on api.list_users.
+--
+-- Body-drift anchor (captured 2026-05-21 pre-write — Phase 1.5):
+--   list_users_for_bulk_assignment    md5: 49e14e620a8b6fc8900cbc018ad5d6bc
+--   list_users_for_role_management    md5: e0805df19e3ae848c51df6a50103d059
+--   list_users_for_schedule_management md5: 3a08d7940a5e322e291ed558426706d8
+--
+-- Idempotent: all three are CREATE OR REPLACE FUNCTION (signature unchanged
+-- -> OID preserved -> existing COMMENT preserved; defensive re-emission
+-- handles future DROP+CREATE).
+--
+-- Session search_path note: the `ltree` parameter type lives in the
+-- `extensions` schema. The function-body `SET search_path` clauses apply
+-- inside the function but NOT during CREATE-time parameter type resolution.
+-- Set the migration-session search_path explicitly so `ltree` resolves.
+-- ============================================================================
+
+SET search_path = public, extensions, pg_temp;
+
+
+-- ============================================================================
+-- 1/3: api.list_users_for_bulk_assignment
+-- ============================================================================
+
+CREATE OR REPLACE FUNCTION api.list_users_for_bulk_assignment(
+  p_role_id     uuid,
+  p_scope_path  ltree,
+  p_search_term text    DEFAULT NULL,
+  p_limit       integer DEFAULT 100,
+  p_offset      integer DEFAULT 0
+)
+RETURNS TABLE(
+  id                  uuid,
+  email               text,
+  display_name        text,
+  is_active           boolean,
+  current_roles       text[],
+  is_already_assigned boolean
+)
+LANGUAGE plpgsql
+SECURITY DEFINER
+SET search_path TO 'public', 'extensions', 'pg_temp'
+AS $function$
+DECLARE
+  v_user_scope extensions.ltree;
+  v_org_id UUID;
+BEGIN
+  -- Get user's scope for permission check
+  v_user_scope := public.get_permission_scope('user.role_assign');
+
+  IF v_user_scope IS NULL THEN
+    RAISE EXCEPTION 'Missing permission: user.role_assign'
+      USING ERRCODE = 'insufficient_privilege';
+  END IF;
+
+  -- Verify requested scope is within user's scope
+  IF NOT (v_user_scope @> p_scope_path) THEN
+    RAISE EXCEPTION 'Requested scope is outside your permission scope'
+      USING ERRCODE = 'insufficient_privilege';
+  END IF;
+
+  -- Get organization ID from scope path (root of path)
+  SELECT o.id INTO v_org_id
+  FROM organizations_projection o
+  WHERE o.path = subpath(p_scope_path, 0, 1)
+    AND o.deleted_at IS NULL;
+
+  IF v_org_id IS NULL THEN
+    RAISE EXCEPTION 'Organization not found for scope path'
+      USING ERRCODE = 'P0002';
+  END IF;
+
+  RETURN QUERY
+  WITH user_current_roles AS (
+    -- Get current role names for each user
+    SELECT
+      ur.user_id,
+      array_agg(DISTINCT r.name ORDER BY r.name) AS role_names
+    FROM user_roles_projection ur
+    JOIN roles_projection r ON r.id = ur.role_id
+    WHERE r.deleted_at IS NULL
+      AND r.is_active = true
+    GROUP BY ur.user_id
+  ),
+  already_assigned AS (
+    -- Users already assigned to this role at this scope
+    SELECT ur.user_id
+    FROM user_roles_projection ur
+    WHERE ur.role_id = p_role_id
+      AND ur.scope_path = p_scope_path
+  )
+  SELECT
+    u.id,
+    u.email::TEXT,
+    COALESCE(u.name, u.email)::TEXT AS display_name,  -- Fallback to email if name is null
+    u.is_active,
+    COALESCE(ucr.role_names, ARRAY[]::TEXT[]) AS current_roles,
+    (aa.user_id IS NOT NULL) AS is_already_assigned
+  FROM users u
+  LEFT JOIN user_current_roles ucr ON ucr.user_id = u.id
+  LEFT JOIN already_assigned aa ON aa.user_id = u.id
+  -- Membership: canonical oracle is users.accessible_organizations (maintained
+  -- by trg_sync_accessible_orgs from user_organizations_projection). Replaces
+  -- the previous u.current_organization_id = v_org_id check, which conflated
+  -- the active-session pointer with organizational membership. See PR #66 and
+  -- migration 20260519233323_fix_list_users_include_roleless.sql for the
+  -- convention origin and the GIN-index rationale (idx_users_accessible_orgs_gin).
+  WHERE u.accessible_organizations @> ARRAY[v_org_id]::uuid[]
+    AND u.deleted_at IS NULL
+    AND (
+      p_search_term IS NULL
+      OR u.name ILIKE '%' || p_search_term || '%'
+      OR u.email ILIKE '%' || p_search_term || '%'
+    )
+  ORDER BY
+    is_already_assigned ASC,  -- Non-assigned first
+    COALESCE(u.name, u.email) ASC
+  LIMIT p_limit
+  OFFSET p_offset;
+END;
+$function$;
+
+-- Edit existing prose to mention accessible_organizations as the membership oracle.
+COMMENT ON FUNCTION api.list_users_for_bulk_assignment(uuid, ltree, text, integer, integer) IS
+$comment$List users in an organization eligible for bulk role assignment to a specific role at a specific scope. Includes current role names per user and an is_already_assigned flag. Membership gated by users.accessible_organizations (the canonical membership oracle, maintained by trg_sync_accessible_orgs from user_organizations_projection). Permission gated by has_permission('user.role_assign') within the requested scope path.
+
+@a4c-rpc-shape: read$comment$;
+
+
+-- ============================================================================
+-- 2/3: api.list_users_for_role_management
+-- ============================================================================
+
+CREATE OR REPLACE FUNCTION api.list_users_for_role_management(
+  p_role_id     uuid,
+  p_scope_path  ltree,
+  p_search_term text    DEFAULT NULL,
+  p_limit       integer DEFAULT 100,
+  p_offset      integer DEFAULT 0
+)
+RETURNS TABLE(
+  id            uuid,
+  email         text,
+  display_name  text,
+  is_active     boolean,
+  current_roles text[],
+  is_assigned   boolean
+)
+LANGUAGE plpgsql
+SECURITY DEFINER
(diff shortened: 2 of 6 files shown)


infrastructure/supabase/supabase/migrations/20260521195657_fix_list_users_sister_functions_membership_gating.sql
```diff
@@ -0,0 +1,428 @@
...
+    is_already_assigned ASC,  -- Non-assigned first
+    COALESCE(u.name, u.email) ASC
+  LIMIT p_limit
+  OFFSET p_offset;
+END;
+$function$;
+
+-- Edit existing prose to mention accessible_organizations as the membership oracle.
+COMMENT ON FUNCTION api.list_users_for_bulk_assignment(uuid, ltree, text, integer, integer) IS
+$comment$List users in an organization eligible for bulk role assignment to a specific role at a specific scope. Includes current role names per user and an is_already_assigned flag. Membership gated by users.accessible_organizations (the canonical membership oracle, maintained by trg_sync_accessible_orgs from user_organizations_projection). Permission gated by has_permission('user.role_assign') within the requested scope path.
```
lars-tice: **Should-fix — COMMENT prose vs body drift.** The COMMENT says "Permission gated by `has_permission(...)`," but the function body uses `public.get_permission_scope("user.role_assign")` followed by `v_user_scope @> p_scope_path` — that is neither `has_permission` nor `has_effective_permission`. It is a scope-extraction + ltree-containment check (see baseline_v4 L7799 for `get_permission_scope` definition — `LIMIT 1` semantics).

This documentation inaccuracy was inherited from the existing pattern, but since we are authoring brand-new prose here (and reissuing on `list_users_for_role_management` below), prefer something like: _"Permission gated by `get_permission_scope("user.role_assign")` returning a scope-path that ltree-contains the requested `p_scope_path`."_ Same edit on the L301 COMMENT for `list_users_for_role_management`, and update the parallel prose in `documentation/architecture/authorization/rbac-architecture.md` (line ~494) for consistency.

Why this matters: the prose is the consumer-facing docstring used by the registry/M3 surface and downstream readers; calling it "has_permission" implies a binary check, hiding the fact that scope-containment is doing real work. (Rule 17 — COMMENT survival OK; the issue is content accuracy, not presence.)

infrastructure/supabase/supabase/migrations/20260521195657_fix_list_users_sister_functions_membership_gating.sql
```diff
@@ -0,0 +1,428 @@
...
+-- -> OID preserved -> existing COMMENT preserved; defensive re-emission
+-- handles future DROP+CREATE).
+--
+-- Session search_path note: the `ltree` parameter type lives in the
+-- `extensions` schema. The function-body `SET search_path` clauses apply
+-- inside the function but NOT during CREATE-time parameter type resolution.
+-- Set the migration-session search_path explicitly so `ltree` resolves.
+-- ============================================================================
+
+SET search_path = public, extensions, pg_temp;
```
lars-tice, line 72: **Should-consider — codify the migration-session `search_path` pattern in `infrastructure/supabase/CLAUDE.md`.** The author flagged this in the PR body and the rationale is sound: function-attribute `SET search_path` does NOT apply during parameter-type resolution at `CREATE OR REPLACE FUNCTION` parse time, only inside the body. Any future migration that uses `ltree`, `vector`, `pg_trgm` types (or anything else under `extensions`) in a parameter list will hit the same `type "ltree" does not exist (SQLSTATE 42704)` deploy failure.

Suggest a short callout in `infrastructure/supabase/CLAUDE.md` under "Supabase CLI Migrations" — three sentences max — mirroring the inline rationale here. This avoids the next contributor losing a deploy cycle to the same gotcha. Worth folding into this PR while the context is fresh, or seeding a tiny follow-up card.

infrastructure/supabase/supabase/migrations/20260521195657_fix_list_users_sister_functions_membership_gating.sql
```diff
@@ -0,0 +1,428 @@
...
+--   two role-functions; has_effective_permission('user.schedule_manage',
+--   <ou_or_org_path>) for schedule_management). The predicate swap broadens
+--   what's enumerable AT a given scope, but the scope itself is still
+--   gated by the existing checks.
+-- - NOT reframing `current_organization_id` semantics — it remains the
+--   active-session pointer for direct-care use.
+-- - NOT addressing super_admin invisibility. Super_admins on dev have
+--   `current_organization_id IS NULL` AND `accessible_organizations IS NULL`.
+--   `NULL::uuid[] @> ARRAY[<uuid>]::uuid[]` returns NULL (treated as
+--   not-matching by WHERE), so super_admins remain excluded post-fix.
```
lars-tice: **Should-consider — super_admin invisibility is acceptable here, but flag the asymmetry with PR #66 explicitly.** PR #66 added an early-return tenancy guard: `IF NOT (has_platform_privilege() OR p_org_id = get_current_org_id()) THEN RETURN; END IF;`. That guard does two things at once: (a) cross-tenant short-circuit, and (b) explicit *platform-privilege bypass* — super_admins are allowed past the tenant check.

These three RPCs do NOT need (a) because the existing `get_permission_scope("user.role_assign") @> p_scope_path` check (and the schedule_management JWT-org check) effectively bound the caller to their own org. But they ALSO do NOT get (b): super_admins remain invisible from their own list-output because `accessible_organizations IS NULL` for them.

That asymmetry — "super_admins are bypassed-into in `list_users` but excluded-from the candidate list in the sister RPCs" — is fine for today (super_admins do not normally hold narrow `role_id`/`schedule_template_id` assignments), but the architectural reasoning should appear in the header so the next reviewer does not re-litigate. Two lines, something like: _"Note: unlike PR #66, no `has_platform_privilege()` bypass is needed at the membership predicate because (1) the existing permission-scope check already bounds callers to their own org, and (2) super_admins do not hold per-scope `role_id`/template assignments — they administer via platform-level paths."_

infrastructure/supabase/supabase/migrations/20260521195657_fix_list_users_sister_functions_membership_gating.sql
```diff
@@ -0,0 +1,428 @@
...
+      WHEN ucs.schedule_template_id IS NOT NULL
+        AND ucs.schedule_template_id <> p_template_id
+      THEN ucs.schedule_name
+      ELSE NULL
+    END AS current_schedule_name
+  FROM users u
+  LEFT JOIN user_current_schedule ucs ON ucs.user_id = u.id
+  LEFT JOIN assigned_to_this_template att ON att.user_id = u.id
+  -- See bulk_assignment comment block above for the membership-oracle rationale.
+  WHERE u.accessible_organizations @> ARRAY[v_org_id]::uuid[]
```
lars-tice, line 381: **Nit — `accessible_organizations` predicate matches sister RPCs (good consistency).** Confirmed for the third surface: same `@> ARRAY[v_org_id]::uuid[]` shape as L180 and L284, no `= ANY` survives anywhere. GIN-friendly per PR #66 convention. The narrower contrast here: this RPC derives `v_org_id` from `get_current_org_id()` (JWT) rather than from `subpath(p_scope_path, 0, 1)` — so the membership query is implicitly tenancy-bounded by the caller`s JWT. That is stricter than the role-function siblings (which can be asked about any scope within the caller`s permission scope, including across orgs if the caller had ever been granted such a scope), and stricter than PR #66`s `list_users` (which took `p_org_id` as a parameter and added an explicit tenancy guard). Worth noting in the header that the three RPCs have *three different* tenancy-derivation mechanisms — they all happen to be sound, but a reader auditing for cross-tenant leakage benefits from that mapping being explicit.

documentation/architecture/authorization/rbac-architecture.md
```diff
@@ -491,21 +491,31 @@ For bulk operations where administrators need to add and remove multiple role as
 
 #### `api.list_users_for_role_management`
 
-Returns all users in the organization with their current assignment status for a specific role.
+Returns all users in the organization with their current assignment status for a specific role at a specific scope. Membership gated by `users.accessible_organizations` (the canonical membership oracle, maintained by `trg_sync_accessible_orgs` from `user_organizations_projection`). Permission gated by `has_permission('user.role_assign')` within the requested scope path.
```
lars-tice: **Should-fix — same content drift as the COMMENT prose in the migration.** This doc now reads "Permission gated by `has_permission(...)`," but the function body uses `get_permission_scope(...)` + ltree-containment. Either keep the doc deliberately abstract ("a scope-bound permission check") or describe the actual mechanism ("`get_permission_scope("user.role_assign")` returning a path that ltree-contains `p_scope_path`"). The former is also fine if the audience is treated as "user of the RPC" not "implementer." Consistency with the COMMENT prose is the only hard requirement.

lars-tice (commented):
# Architectural Review — PR #67

**Verdict: APPROVE WITH IN-PR FIXES**

This is a clean, surgical fix that correctly applies PR #66's `accessible_organizations @> ARRAY[<uuid>]::uuid[]` convention to all three sister RPCs. The body diff is exactly one predicate per RPC; the permission gates, scope-derivation logic, and return shapes are preserved verbatim. Signatures unchanged → `database.types.ts` / `rpc-registry.generated.ts` byte-identical (verified). Rule 17 (RPC shape tag) is satisfied across all three. The transactional `BEGIN; ... ROLLBACK;` smoke harness is a strict upgrade over PR #66's against-live-state smoke.

## Per-finding table

| # | Severity | Finding | Action |
|---|---|---|---|
| 1 | Should-fix | COMMENT prose says `has_permission('user.role_assign')` but body uses `get_permission_scope(...) + ltree @>`. Same drift in `rbac-architecture.md` line 494. | Tighten prose in this PR — three brand-new comment blocks are being authored here, no reason to inherit the inaccuracy. |
| 2 | Should-consider | Super_admin invisibility is acceptable but the *asymmetry* with PR #66's `has_platform_privilege()` bypass is undocumented. | Two-line header addition explaining why this RPC class doesn't need the bypass (caller's scope check is org-bounded; super_admins don't hold narrow assignments). |
| 3 | Should-consider | Migration-session `SET search_path` workaround is a real gotcha — any future `ltree`/`vector`/`pg_trgm` parameter migration will hit it. Author flagged in PR body. | Codify in `infrastructure/supabase/CLAUDE.md` — three-sentence callout under "Supabase CLI Migrations." Fold into this PR while context is fresh, or seed a tiny follow-up card. |
| 4 | Nit | Three sister RPCs derive `v_org_id` three different ways (subpath-of-scope-path, JWT). All sound; would help an auditor to map them in the migration header. | Optional — header addition. |
| 5 | Nit | The `list_users_for_bulk_assignment` baseline COMMENT had prose but no shape tag — meaning the rpc-registry was picking it up via some other path (likely a later migration not surfaced in the diff). New COMMENT correctly carries the tag. | None — defensive reissue is the right call. |

## Conformance check (binding rubric)

| Rule | Status | Notes |
|---|---|---|
| Rule 1 — CLI-generated timestamp | PASS | `20260521195657_*.sql` is CLI-generated. |
| Rule 2 — Idempotency | PASS | All three are `CREATE OR REPLACE FUNCTION`, signatures unchanged → OID preserved. |
| Rule 4 — Frontend via `api.*` only | PASS (preserved) | These ARE `api.*` RPCs; no consumer changes. |
| Rule 15 — `database.types.ts` regen | PASS — N/A | Signatures unchanged, no regen needed (verified `git diff main..HEAD` on both files is empty). |
| Rule 16 — RAISE EXCEPTION no identifiers | PASS | `Missing permission: user.role_assign` is a fixed enum value, not an identifier. No `%` interpolations in the migration. |
| Rule 17 — `@a4c-rpc-shape` COMMENT | PASS | All three carry `@a4c-rpc-shape: read`; defensive reissue handles future DROP+CREATE. |
| Rule 19 — Edge Function cross-schema | N/A | SQL-only migration. |
| Function-attribute `SET search_path` | PASS | All three carry `SET search_path TO 'public', 'extensions', 'pg_temp'`. |
| Migration-session `SET search_path` | PASS | Present at top (L90). |
| Docs frontmatter `last_updated` | PASS | Both `rbac-architecture.md` and `users.md` bumped to 2026-05-21. |
| Predicate shape (no `= ANY` survives) | PASS | All three sites use `@> ARRAY[v_org_id]::uuid[]`. |
| Permission-gate preservation | PASS | Body diff vs baseline: only the membership predicate changed. |
| `COUNT(*) OVER ()` dedup omitted | PASS — correct call | These RPCs don't return `total_count`. Confirmed by reading the RETURNS TABLE signatures. |
| Smoke verification | PASS | Transactional `BEGIN; ... ROLLBACK;` is methodologically stronger than PR #66's against-live smoke. |

## Architecture review checklist

- **CQRS Standards** — PASS. Reads from projections via `api.*` RPC, no PostgREST embedding.
- **Naming Conventions** — PASS. `list_users_for_*` family naming preserved; predicate naming follows PR #66 convention.
- **Design Patterns** — PASS. Convention-replication pattern correctly applied across three surfaces.
- **data-testid Attributes** — N/A (no UI changes).
- **AsyncAPI Event Registration** — N/A (no new events).
- **Type Generation — No Anonymous Types** — PASS. RPC return shapes are explicit `RETURNS TABLE(...)`; generated TS types unchanged.
- **Observability, Tracing & Monitoring** — PASS. `RAISE EXCEPTION ... USING ERRCODE` paths preserved; trigger-tier `processing_error` capture path unaffected.
- **Error Surfacing to UI** — PASS. RPC error shape unchanged; existing `apiRpc<T>` consumer path masks at the SDK boundary.

## Punch list (before merge)

- [ ] Tighten COMMENT prose on `list_users_for_bulk_assignment` (L197) and `list_users_for_role_management` (L301) to accurately describe the `get_permission_scope` + ltree-containment gate.
- [ ] Mirror the same prose fix in `documentation/architecture/authorization/rbac-architecture.md` line 494.
- [ ] (Optional but cheap) Add two-line header note explaining why no `has_platform_privilege()` bypass is needed here.
- [ ] (Optional but cheap) Codify the `SET search_path` migration-session pattern in `infrastructure/supabase/CLAUDE.md`.

No blockers. The fixes above are P3 cleanup; the load-bearing predicate change is correct.

---

lars-tice:
## Review-response summary — all 5 findings addressed + Item #4 escalated via threat-model audit

Pushed \`1207ca46\` to address every finding from the 2026-05-21 architectural review. Item #4 was escalated by the user from "document the divergence in the migration header" to "unify the approach," then grounded by a parallel threat-model audit (three Explore agents) that surfaced **cross-tenant-grant forward-compatibility** as the load-bearing concern.

### Per-finding resolution

**Finding #1 (Should-fix) — COMMENT prose drift** ✅
- All three sister COMMENTs updated to the architect-Finding-#5 exact wording: \`"Permission gated by has_effective_permission('<perm>', <path>) (verifies the caller holds the permission at a scope that contains the requested path)."\`
- Mirror fix in \`documentation/architecture/authorization/rbac-architecture.md\` L494.

**Finding #2 (Should-consider) — codify SET search_path gotcha** ✅
- New subsection in \`infrastructure/supabase/CLAUDE.md\` under "Supabase CLI Migrations": "Migration-session SET search_path gotcha" (3 sentences, references PR #67 as discovery).

**Finding #3 (Should-consider) — super_admin asymmetry** ✅
- Implicitly resolved by Finding #4's normalization: all three sister RPCs now use the same \`has_effective_permission\` shape, so the asymmetry framing is moot. Header drops the super_admin block; the relevant context lives in the new CLAUDE.md "list_users* family pattern" subsection.

**Finding #4 (Nit → escalated)** ✅
The two role-functions refactored from the two-step pattern to a single \`has_effective_permission\` call:

\`\`\`sql
-- BEFORE (two-step):
v_user_scope := public.get_permission_scope('user.role_assign');
IF v_user_scope IS NULL THEN
  RAISE EXCEPTION 'Missing permission: user.role_assign' USING ERRCODE = 'insufficient_privilege';
END IF;
IF NOT (v_user_scope @> p_scope_path) THEN
  RAISE EXCEPTION 'Requested scope is outside your permission scope' USING ERRCODE = 'insufficient_privilege';
END IF;

-- AFTER (single-step; semantically equivalent today, strictly more correct under cross-tenant grants):
IF NOT public.has_effective_permission('user.role_assign', p_scope_path) THEN
  RAISE EXCEPTION 'Missing permission: user.role_assign' USING ERRCODE = 'insufficient_privilege';
END IF;
\`\`\`

All three sister RPCs now share the same permission-check helper + the same membership predicate. \`list_users_for_schedule_management\` was already on this pattern; no change to it.

**Why strictly more correct, not just equivalent**: \`get_permission_scope\` does \`LIMIT 1\`; \`has_effective_permission\` does \`EXISTS\`. They're observationally identical today because \`compute_effective_permissions\` (baseline_v4:6932-6985) ends in \`DISTINCT ON (permission_name)\` — JWT carries at most one entry per permission. **If that invariant breaks** (cross-tenant grants producing multiple JWT entries for the same permission at distinct scopes), the LIMIT-1 pattern silently picks first and may miss a valid match; EXISTS correctly ORs across all entries. Tripwire documented in the migration header.

**Finding #5 (Nit) — defensive shape-tag reissue**: no action needed.

### Plan-review architect findings (folded before execution)

The plan itself went through architect review. 4 substantive findings (Must-fix #1, Must-fix #2, Should-consider #3, Should-consider #4) plus nits, all folded into the plan before execution:

- **Must-fix #1**: equivalence claim strengthened with \`DISTINCT ON\` tripwire (now in migration header + CLAUDE.md)
- **Must-fix #2**: **Four-site distribution memorialized**. Grep on baseline_v4 finds the legacy two-step pattern in 4 sites:

  | Line | Function | Class | Touched here? |
  |---|---|---|---|
  | L362 | \`api.bulk_assign_role\` | Mutation | No |
  | L4705 | \`api.list_users_for_bulk_assignment\` | Visibility | **Yes** |
  | L4793 | \`api.list_users_for_role_management\` | Visibility | **Yes** |
  | L5571 | \`api.sync_role_assignments\` | Mutation | No |

  Scope decision explicit in migration header: this PR normalizes the **visibility-class** siblings only. The two mutation siblings (\`bulk_assign_role\`, \`sync_role_assignments\`) stay on the legacy pattern; a future card may extend normalization. Mutations carry side-effect risk warranting a focused diff.

- **Should-consider #3**: normalized-pattern documentation relocated from migration header to \`infrastructure/supabase/CLAUDE.md\` as new "list_users* family pattern — three-step skeleton" subsection. Header now compact.
- **Should-consider #4**: dropped standalone threat-model memory file from the plan; will fold into future \`pr-67-close-out.md\` on merge (matches \`pr-63-\`/\`pr-66-close-out.md\` conventions).

### Deferred concern (captured for future card)

The threat-model audit's central finding: **PR #66's existing tenancy guard would block legitimate cross-tenant grant scenarios** when partner-grant work activates. Partner consultants with valid grants have \`accessible_organizations @> [home, providerA]\` but JWT \`org_id\` stays at home org — PR #66's \`IF NOT (... OR p_org_id = get_current_org_id()) ...\` guard returns empty for them. The three sister RPCs' scope-bound permission checks are forward-compatible.

Deferred per user choice. Will be captured in \`pr-67-close-out.md\` post-merge under "Deferred: cross-tenant-grant audit" subsection, with a cross-reference from \`pr-66-close-out.md\` on next groom. Future card trigger: when work activates per \`dev/active/sub-tenant-admin-design/\` + \`documentation/architecture/data/provider-partners-architecture.md\`.

### Verification

- \`supabase migration repair --status reverted 20260521195657\` + \`supabase db push --linked\` (per \`pr-63-close-out\` migration-repair pattern, safe on feature branches not yet merged)
- All 3 sister RPC bodies now: \`uses_has_effective_permission=true\`, \`uses_get_permission_scope=false\`, \`has_old_error_message=false\`
- Re-ran PR #67 Phase 3 transactional smoke harness — identical pass results pre/post refactor (confirms semantic equivalence today)
- Frontend gates: typecheck + lint + docs:check + build all green

Ready for re-review.

lars-tice (commented):
# Re-review of `1207ca46` — APPROVE (posted as comment — cannot self-approve)

**Verdict**: APPROVE. All 5 findings from `59a631a7` review verified resolved. The F4 escalation (single-call `has_effective_permission` refactor) is correct, well-reasoned, and strictly more correct under future cross-tenant grants. CI green (Validate PL/pgSQL Functions pass, Verify rpc-registry pass).

## Per-finding verification

### F1 (Should-fix — COMMENT prose drift): RESOLVED

Exact wording present in all three `COMMENT ON FUNCTION` blocks:
- L173–174 `list_users_for_bulk_assignment` — verbatim
- L272–273 `list_users_for_role_management` — verbatim
- L396–397 `list_users_for_schedule_management` — adapted form `has_effective_permission('user.schedule_manage', <ou_path OR org_path>)` (correct for the COALESCE'd path expression in this RPC; the prose accurately describes the path being checked)

`documentation/architecture/authorization/rbac-architecture.md:496` carries the matching mirror sentence. Done correctly.

### F2 (Should-consider — Migration-session SET search_path codification): RESOLVED

New subsection "Migration-session `SET search_path` gotcha (extension-typed parameters)" at `infrastructure/supabase/CLAUDE.md` (under "Supabase CLI Migrations"). Accurately states:
- Function-attribute `SET search_path` doesn't apply during CREATE-time parameter parsing
- Extension-typed parameters (`ltree`, `vector`, `pg_trgm`) trip this
- Failure mode: `type "<type>" does not exist (SQLSTATE 42704)`
- Mitigation: session-level `SET search_path = public, extensions, pg_temp;` at top of migration
- PR #67 cited as discovery

Reusable for any future migration that adds an `ltree` parameter.

### F3 (Should-consider — Pattern docs relocated): RESOLVED

Migration header (L1–70) is now compact: PR-66 precedent reference, four-site distribution table (L26–42), DISTINCT ON tripwire (L44–55), search_path note (L57–60), body-drift anchor hashes (L62–65), idempotency note (L67–69), and pointer to CLAUDE.md. Verbose content moved to the new "`list_users*` family pattern — three-step skeleton" subsection in `infrastructure/supabase/CLAUDE.md`. The subsection accurately captures the three-step skeleton (permission gate → v_org_id derivation → membership predicate), the signature-dependent v_org_id variant, the canonical membership-oracle rule, the two-step → one-step normalization rationale, and the four-site distribution audit.

### F4 (Nit → ESCALATED to unify): RESOLVED — equivalence claim verified

- L107 `list_users_for_bulk_assignment` and L211 `list_users_for_role_management` now both use:
  ```sql
  IF NOT public.has_effective_permission('user.role_assign', p_scope_path) THEN
    RAISE EXCEPTION 'Missing permission: user.role_assign'
      USING ERRCODE = 'insufficient_privilege';
  END IF;
  ```
- L328 `list_users_for_schedule_management` was already on this pattern and is unchanged in structure (still has its multi-path COALESCE'd scope expression). Author claim verified.

**Equivalence claim verification** (read `compute_effective_permissions` and both helpers):
- `compute_effective_permissions` (baseline L6976–6982) ends in `SELECT DISTINCT ON (permission_name) ... ORDER BY permission_name, nlevel(scope_path) ASC` — confirms JWT carries at most one entry per permission_name today
- `get_permission_scope` (baseline L7799–7812) does `LIMIT 1` on JWT effective_permissions
- `has_effective_permission` (baseline L9827–9842) does `EXISTS ... WHERE ep->>'p' = p_permission AND (ep->>'s')::ltree @> p_target_path` — correctly ORs across all matching entries

Confirmed: observationally identical TODAY (DISTINCT ON guarantees single entry → LIMIT 1 picks the only one). If cross-tenant grants break the DISTINCT ON invariant by producing multiple per-permission JWT entries at distinct scopes, the LIMIT-1 path silently misses while EXISTS correctly evaluates. **The refactor is strictly more correct, as claimed.**

**Four-site distribution verified** (grep `v_user_scope := public.get_permission_scope` against baseline_v4):
- L353  `api.bulk_assign_role` (mutation — out of scope, intentional)
- L4696 `api.list_users_for_bulk_assignment` (visibility — normalized in this PR)
- L4784 `api.list_users_for_role_management` (visibility — normalized in this PR)
- L5562 `api.sync_role_assignments` (mutation — out of scope, intentional)

Exactly four sites. Author's header line refs (L362/L4705/L4793/L5571) point to the `'Requested scope is outside your permission scope'` error line within each block — same hits.

**Super_admin compatibility verified**: `custom_access_token_hook` (baseline L7123–7137) grants super_admin ALL permissions at scope `''` (empty ltree = global root). Since empty ltree `@>` contains every path, `has_effective_permission('user.role_assign', <anything>)` returns TRUE for any super_admin. No explicit `has_platform_privilege()` short-circuit needed — the permission gate handles super_admin naturally. Semantically equivalent to PR #66's tenancy guard for the super_admin axis.

**Error message + ERRCODE**: `'Missing permission: user.role_assign'` + `ERRCODE = 'insufficient_privilege'` matches the existing pattern (e.g., baseline L356, L4699, L4787, L5565). Rule 16 compliant — no PII interpolation.

### F5 (Nit — defensive shape-tag reissue): RESOLVED

All three COMMENT blocks carry `@a4c-rpc-shape: read` (L176, L275, L399). CI's `Verify rpc-registry.generated.ts matches migration state` passes — shape tags are sound.

## Cross-cutting checks

- **Consistency with PR #66's tenancy guard**: The `has_effective_permission` form supersedes the simpler `has_platform_privilege() OR p_org_id = get_current_org_id()` guard for role-management RPCs because the scope-path signature carries a finer-grained authorization target. For super_admin: empty-scope global grant ensures the permission gate passes. For cross-tenant non-super_admins: their JWT effective_permissions won't contain a `user.role_assign` entry scoped to the target tenant's path, so the gate correctly rejects. Sound architectural fit.
- **No PII leakage**: All RAISE EXCEPTION strings are static.
- **Idempotency**: `CREATE OR REPLACE FUNCTION` with unchanged signatures → OID preserved → existing COMMENT survives. Defensive re-emission still correct for future DROP+CREATE scenarios.
- **CI status**: Validate PL/pgSQL Functions PASS (19s); Verify rpc-registry PASS (2m35s); Deploy Migrations skipping (expected on PR-open).

## New findings

None. The second commit improves correctness without introducing new risk.

## Punch list

None. Ready to merge.

---
*Re-review of head SHA `1207ca465916fbdaf0630f3bf89c26e9a399edfc` against findings from `59a631a7` review (2026-05-21).*

PR #78, opened 2026-06-15
refactor(users): remove per-user org-override phones & addresses (global-only)

lars-tice:
## Summary

Removes the per-user **org-override** contact tables `user_org_phone_overrides` and `user_org_address_overrides` (keyed `user_id`+`org_id`). All user phones/addresses are now **global** (`user_phones`/`user_addresses`).

These overrides were a structural-symmetry "hybrid scope" feature (user-management Decision #13, 2024-12-31) with **no documented use case** — only **1 unused row** existed system-wide. The org-level HQ contact subsystem (`phones_projection`/`addresses_projection`/`contacts_projection` + `organization_*` junctions, keyed `organization_id`) is a **separate subsystem and is untouched**.

**Trigger:** the notification-prefs SMS picker offered org-scoped phones, but `user_notification_preferences_projection.sms_phone_id` FK → `user_phones(id)` (global only) → FK violation on save (`Could not save notification preferences`). Removing org-override phones fixes it structurally — the picker only ever sees global phones now.

## Changes

**Backend** (`migration 20260615175954`): simplify 5 `api` RPCs (`add`/`update`/`remove`/`get_user_phones`/`get_user_sms_phones`) + 3 `public` helpers + 6 handlers (drop the org/`p_org_id` branch); re-issue 2 stale RPC comments (tags verbatim); drop 6 RLS policies, 5 indexes, 2 tables; inverted pitfall-#8 assertion (tables + refs gone). `CREATE OR REPLACE` preserves OIDs + `@a4c-rpc-shape`/reachability tags; **vestigial `p_org_id` params kept (always NULL)** to avoid a signature/regen break.

**Frontend:** drop `orgId`/`source` from `UserPhone`/`UserAddress` + request types; remove the org-override UI (global/org split, `isOrgOverride` forms, org badges); harden `NotificationPreferencesForm` SMS auto-select (empty → `null` + `log.warn`); hand-trim the two dropped-table blocks from **both** `database.types.ts` copies (RPC sigs unchanged → no `Functions` change).

**Docs:** org-HQ-vs-user distinction note in `organization-management-architecture.md`; Decision #13 reversal in `user-management-context.md`; AsyncAPI `org_id` field deprecated-in-description (kept optional, retained for immutable event history).

## Verification

- **Dev (applied via Mgmt API):** both tables gone, consumer audit empty, `add_user_phone` + **notification-prefs save** (the original bug) + `user.address.added` (org-less) all succeed; test rows torn down.
- **Gates:** `npm run typecheck` (frontend + workflows) ✅, `lint` ✅, `vitest src/services/users/__tests__/` 41/41 ✅, `docs:check` ✅. `supabase db lint` + `rpc-registry-sync` validated by CI.
- **Architect-reviewed** APPROVE WITH IN-PR FIXES (S1/S2/S3/N1/N2 folded).

## Ordering / notes

- ⚠️ **Cross-card:** edited `dev/active/handle-user-deleted-cascade-cleanup-projections/plan.md` to drop the two override-table DELETEs (would otherwise deploy a handler referencing dropped tables). **This PR lands first.**
- **N1 (parked):** `dev/parked/handler-code-generation/handler-code-generation-context.md:111` references `user_org_phone_overrides` as a codegen example — parked/low-risk, flagged so a future thaw doesn't regenerate a handler for a dropped table.
- **Parked follow-ups** (separate cards): (1) org-HQ→user contact copy/mirror affordance for solo/family providers; (2) phone-number E.164 normalization + per-user dedup (must NOT be per-org/global — would forbid legitimately shared family lines).


infrastructure/supabase/contracts/asyncapi/domains/user.yaml
```diff
@@ -604,7 +604,7 @@ components:
           type: string
           format: uuid
           nullable: true
-          description: If set, this is an org-specific override address
+          description: "Deprecated (org-override removed 2026-06): always null; retained for event-history compatibility."
```
lars-tice, line 607: **N2 (non-blocking) — regenerate the AsyncAPI bundle + sweep `invitation.yaml`.**

This source edit is correct (deprecation note, field kept optional/nullable for immutable event history). But the generated bundle was not regenerated from it:

- `infrastructure/supabase/contracts/asyncapi-bundled.yaml` still carries the OLD `"If set, this is an org-specific override ..."` prose at ~10 sites (e.g. L891, L1299, L19150). Run `cd infrastructure/supabase/contracts && npm run generate:types` and commit the regenerated bundle + `generated-events.ts` so the wire contract matches the source.
- `infrastructure/supabase/contracts/asyncapi/domains/invitation.yaml:249` still describes `phone_id` as coming "from user_phones or `user_org_phone_overrides`" — update that description to drop the dropped table.

Non-blocking: descriptions only; `org_id`/`phone_id` stay nullable so event-history compatibility is preserved.

lars-tice (commented):
## Architectural Review — PR #78 (remove per-user org-override phones & addresses)

**Verdict: APPROVE WITH IN-PR FIXES** (both findings are non-blocking N-level; the migration itself is correct and CI is green).

Reviewed against the codified pitfalls in `infrastructure/supabase/CLAUDE.md` (esp. #6 deployed-body diff, #8 column-drift assertion), the RPC shape/reachability registries, and the CQRS/event-handler conventions. Verified live against the deployed DB (`tmrjlswbsxmbglmaclxu`) via the Mgmt API SQL endpoint.

### What I verified (all PASS)

- **Pitfall #6 (no load-bearing semantics dropped on CREATE OR REPLACE).** Diffed each of the 5 RPC + 3 helper + 6 handler new bodies against the *deployed* pre-migration bodies (baseline_v4 → `20260423232531` → hotfix `20260610200218`). The collapse of the two-branch (`IF p_org_id IS NULL ... ELSE`) logic to the global-only branch preserves every retained invariant: three-tier auth gate, metadata + optional `reason`, `emit_domain_event` payload (incl. vestigial `org_id`), and **Pattern A v2 BOTH-checks read-back** (`IF v_phone IS NULL`/`IF NOT FOUND` + post-read `processing_error` check, no `RAISE EXCEPTION`). `update_user_phone` correctly keeps `row_to_json(v_row)::jsonb` (global `user_phones` carries `is_primary`, so the hotfix's org-branch `|| {is_primary:false}` shim is correctly dropped, not lost).
- **OID / shape-comment / reachability-tag preservation.** Signatures unchanged (vestigial `p_org_id`/`p_organization_id` retained, always NULL) → `CREATE OR REPLACE` preserves OIDs. Confirmed deployed comments for all 5 RPCs still carry `@a4c-rpc-shape` + `@a4c-bucket`/`@a4c-consultant-callable`/`@a4c-phase-target` verbatim. Only the 2 RPCs with stale override prose (`add_user_phone`, `get_user_phones`) are re-issued; tag lines byte-preserved. CI `rpc-registry-sync` + `reachability-matrix-sync` both pass.
- **Inverted pitfall-#8 assertion is correct.** The DO-block asserts both tables are gone (`to_regclass IS NULL`) AND no `public`/`api` function body still ILIKEs either table name. Live re-run of that exact predicate against the deployed DB returns `[]` (zero residual refs) and both `to_regclass` are NULL. Sound inversion of the column-existence pattern.
- **Org-level HQ subsystem genuinely untouched.** `phones_projection`, `addresses_projection`, `organization_phones`, `organization_addresses` all still present on the deployed DB; migration touches none of them. The new scope-note in `organization-management-architecture.md` correctly draws the org-keyed-vs-user-keyed boundary.
- **RLS / index / table drops have no remaining referents.** 6 RLS policies + 5 non-PK indexes + 2 tables dropped; zero **inbound** FKs (the `user_notification_preferences_projection.sms_phone_id` FK targets `user_phones`, never the override tables — which is precisely the structural fix). Handler reference files under `handlers/user/` updated to match deployed bodies.
- **Frontend / type surface.** `UserPhone`/`UserAddress` + request types drop `orgId`/`source` cleanly; both `database.types.ts` copies hand-trimmed in lockstep (118 lines each); `tsc --noEmit` clean; query/command services drop the `.source`/`.orgId` mapping and pass `p_org_id: null`. `NotificationPreferencesForm` SMS auto-select now falls through to `null` (never an undefined phone) with a `log.warn` — this is the correct structural fix for the original FK-violation bug and surfaces an observable signal.
- **Cross-card ordering.** `handle-user-deleted-cascade-cleanup-projections/plan.md` correctly drops the two override-table DELETEs from both the inventory table and the cascade SQL, with a dated guard note. #78-lands-first ordering is respected.

### Findings

**N1 (non-blocking) — the anti-drift contract test is silently no-op'ing on `add_user_phone`/`update_user_phone`.**
`frontend/src/services/users/__tests__/UserRpcContract.test.ts` extracts "the last CREATE OR REPLACE" by a regex terminated on `^\$\$;$` (`extractLastFunctionBody`, ~L62). But the hotfix `20260610200218` and this PR's `20260615175954` both terminate these functions with `$function$;`, **not** `$$;`. So the regex skips both and matches the **two-versions-stale** body in `20260423232531` — which still contains `IF p_org_id IS NULL` and `FROM user_org_phone_overrides`. Result: the suite is green (11/11) but its `'branches on p_org_id ...'` assertion (L102-104) is now validating **dropped objects in dead SQL**, giving false confidence. This defect predates #78 (introduced by the #77 hotfix's terminator), so it does not block this PR — but since #78 is what makes those assertions reference non-existent tables, fix it here: (a) broaden the terminator regex to `(\$\$|\$function\$);`, and (b) update the now-inverted assertions to expect global-only (`FROM user_phones`, no `user_org_phone_overrides`, no `IF p_org_id IS NULL`). Consider a guard test asserting the matched body's source migration is the newest one containing the symbol.

**N2 (non-blocking) — regenerate the AsyncAPI bundle + sweep `invitation.yaml`.** (also posted inline on `user.yaml`)
The `user.yaml` source edits are correct, but `infrastructure/supabase/contracts/asyncapi-bundled.yaml` was not regenerated — it still carries the old `"If set, this is an org-specific override ..."` prose at ~10 sites. And `infrastructure/supabase/contracts/asyncapi/domains/invitation.yaml:249` still describes `phone_id` as sourced "from user_phones or `user_org_phone_overrides`". Run `cd infrastructure/supabase/contracts && npm run generate:types`, fix the `invitation.yaml` description, and commit the regenerated bundle + `generated-events.ts`. Descriptions-only; fields stay nullable so immutable event-history compatibility is preserved.

### Mandatory Review Checklist

- **CQRS Standards** — PASS. Commands emit events via `emit_domain_event`; handlers own projection writes; reads go through `api.*` RPCs. No direct projection writes introduced.
- **Naming Conventions** — PASS. `p_org_id` vestigial-param convention documented inline; handler/router names unchanged; deprecation prose dated.
- **Design Patterns** — PASS. Pattern A v2 read-back preserved; the "retain vestigial param to keep OID + shape comment" approach is the right call over DROP+CREATE+re-tag (avoids registry churn). Appropriately scoped, not over-engineered.
- **data-testid** — N/A / PASS. No new interactive UI; `NotificationPreferencesForm` edits are logic + className-only.
- **AsyncAPI Event Registration** — NEEDS ATTENTION (N2). Source `user.yaml` updated; bundle + `invitation.yaml` not. No event added/removed; `org_id` retained nullable for history. Remediation above.
- **Type Generation — No Anonymous Types** — PASS. Named `UserPhone`/`UserAddress` + request types updated; both `database.types.ts` copies regenerated/trimmed in lockstep; typecheck clean.
- **Observability / Tracing** — PASS. Event metadata (`user_id`, optional `reason`, `source`) preserved on every emit; new `log.warn` adds a previously-silent signal on the SMS-no-phone path.
- **Error Surfacing to UI** — PASS. Pattern A v2 envelope (`success:false` + `processing_error`) intact end-to-end; the FK-violation "Could not save notification preferences" path is now structurally unreachable.

### Decision Record (abbrev.)
Reversal of user-management Decision #13 (2024-12-31 structural-symmetry hybrid-scope) as YAGNI: 1 unused row, no use case, and the source of a live FK-violation bug. Removal is symmetric (tables + RLS + indexes + 14 consumers + types + docs + cross-card) and verified on dev. Org-level HQ contact subsystem explicitly out of scope and confirmed untouched.

lars-tice:
✅ **Both architect findings folded in** (commit `24dac975`):

**N1** — `UserRpcContract.test.ts`: broadened `extractLastFunctionBody` terminator to `^$(?:function)?$;$` so it matches the `$function$;`-tagged bodies, and inverted the `add_user_phone` assertion to global-only (`FROM user_phones`, asserts **no** `IF p_org_id IS NULL` and **no** `user_org_phone_overrides`). The inverted assertions passing confirms the matcher now locks onto the fresh global-only body, not the stale `20260423232531` one. 11/11 tests green.

**N2** — fixed `invitation.yaml` `SmsNotificationConfig.phone_id` description (dropped the `user_org_phone_overrides` reference) and regenerated `asyncapi-bundled.yaml` via `npm run generate:types`. Bundle now has **0** `org-specific override` / `user_org_phone_overrides` sites. Descriptions-only → `generated-events.ts` unchanged (no TS interface drift), so no frontend type copy required.

Verdict stays **APPROVE** — both were N-level, now resolved.

PR #74, opened 2026-06-09
fix(authz): align cross_tenant_access_grants_projection schema with deployed handler

lars-tice:
## Summary

Hotfix for a **baseline_v4 schema-vs-handler column-name mismatch defect** surfaced during Phase 2 UAT execution (2026-06-09). Probe L8 (`api.revoke_access_grant` happy path) returned `PROCESSING_FAILED` with `column \"revocation_reason\" of relation \"cross_tenant_access_grants_projection\" does not exist`. Audit of all router arms revealed **3 of 5 grant lifecycle arms write to columns that don't exist** on the projection.

| Arm | Handler writes | Schema had | Status pre-hotfix |
|---|---|---|---|
| `revoked` | `revocation_reason`, `revocation_details` | `revoked_reason`, no `_details` | ❌ every revoke = `PROCESSING_FAILED` |
| `expired` | `expired_at`, `expiration_type` | neither column | ❌ all expirations broken |
| `reactivated` | `resolution_details` | `reactivation_notes` (different name) | ❌ all reactivations broken |
| `suspended` | all 5 cols | ✅ aligned | works |
| `policy_override` | `permissions` jsonb only | ✅ aligned | works |

## Why this surfaced now

Defect predates Phase 2 — handler was built against a different schema shape than what landed in `baseline_v4`. Surfaced now because Phase 2's cascade-revoke (`terminate_var_partnership` Step 13) is the first production path that emits `access_grant.revoked` at scale. Phase 2 architect review focused on the cascade-FIRST HIPAA invariant (preserved correctly) and Pattern A v2 envelope shape (preserved correctly) — neither line of review opens the handler arms' UPDATE column lists against the projection schema.

## Direction chosen

Per AsyncAPI contract + handler review: handler / AsyncAPI / event_data all use `revocation_*` / `expired_*` / `resolution_*` naming. The projection schema is the outlier. **Aligning the schema is the minimal-blast-radius fix**: no handler changes, no AsyncAPI changes, no event_data changes.

## Migration body (~205 lines)

- **1.1**: rename `revoked_reason` → `revocation_reason` (idempotent via `information_schema` lookups; `RENAME COLUMN IF EXISTS` doesn't exist in PG)
- **1.2**: add `revocation_details text` (idempotent)
- **1.3**: add `expired_at timestamptz` (idempotent)
- **1.4**: add `expiration_type text` (idempotent)
- **1.5**: rename `reactivation_notes` → `resolution_details` (idempotent)
[...]

PR #81, opened 2026-06-22
fix(events): handle_user_deleted cascade-cleans FK-dependent projections

lars-tice:
## Problem

`handle_user_deleted` only tombstoned the user (`deleted_at`/`is_active`); it left stale rows in membership/identity projections. Soft-deleted users still resolved as members (e.g. `check_user_org_membership`), and consumer RPCs needed ad-hoc `WHERE deleted_at IS NULL` filters. Dev had **9 stale rows** across 2 users — this blocked PR #64 UAT T4.

## Fix (two layers — architect-reviewed, APPROVE WITH IN-PR FIXES)

1. **BEFORE handler** `handle_user_deleted` — after the preserved COALESCE tombstone + `IF NOT FOUND` guard, hard-DELETE the 7 membership/identity projections + NULL the `contacts_projection` user-linkage. **Order is load-bearing**: tombstone first, so the `accessible_organizations` recompute that fires on `user_organizations_projection` DELETE self-guards (`deleted_at IS NULL`) to a no-op. Audit-reference columns preserved.
2. **AFTER-INSERT trigger** `emit_grant_revocations_on_user_deleted` — emits `access_grant.revoked` per active grant of the deleted consultant. This is the architect's **F1** fix: keeping it out of the BEFORE handler means a failed revoke surfaces as a retryable `processing_error` instead of being buried in the inner event and returning a silent `{success:true}` to `api.delete_user`'s outer read-back.

**Decisions**: revoke (not transfer); no `auth.admin.deleteUser` cascade; contacts NULL-link only (no PII clear); backfill via direct DELETEs (all 3 arms); keep consumer filters (defense-in-depth).

**F2**: automated revoke emits `revoked_by=NULL` → made `AccessGrantRevokedData.revoked_by` nullable. Note: `rbac.yaml` is the **wired** source (its `AccessGrantRevoked` message is what `asyncapi.yaml` references); `access_grant.yaml` carries a parallel **unwired** duplicate of the schema, aligned here — consolidating the two is a separate cleanup.

Pitfall #6 (deployed body preserved verbatim), #8 (column-existence assertion), #4 N/A. Replay-safe (DELETE/NULL idempotent).

## Verification (dev; transactional simulate-JWT, `BEGIN…ROLLBACK`, zero pollution)

| Check | Result |
|-------|--------|
| Backfill | 9 stale rows → **0** |
| Handler cascade | all 7 projections cleaned + contacts NULL + tombstone, no `processing_error` |
| Grant-revoke arm (synthesized) | `status=revoked`, `revoked_by IS NULL`, reason correct, **1 event, no `processing_error`** |
| Replay | 2nd `user.deleted` is a clean no-op (0 errors) |
| Cleanliness | 0 leaked grants, real user rows intact (all rolled back) |

PR #64 UAT T4 is now runnable as designed (soft-deleted user → `check_user_org_membership` 0 → … → `not_found`).

## Scope

No `database.types.ts` regen (handler + trigger bodies only). `generated-events.ts` regenerated for the contract change (contracts copy committed; frontend copy is gitignored, regenerated by the build).

### DIFF
--- a/infrastructure/supabase/handlers/trigger/emit_grant_revocations_on_user_deleted.sql
+++ b/infrastructure/supabase/handlers/trigger/emit_grant_revocations_on_user_deleted.sql
@@ -1,0 +1,53 @@
+-- AFTER-INSERT trigger fn on public.domain_events WHEN (NEW.event_type = 'user.deleted').
+-- Revokes active cross-tenant grants whose consultant_user_id is the deleted user,
+-- by emitting one access_grant.revoked event per grant. Event-chaining lives in an
+-- AFTER-INSERT trigger (NOT a nested emit from the BEFORE handler handle_user_deleted):
+-- a nested emit would bury an inner failure in the inner event's processing_error,
+-- which api.delete_user's outer Pattern-A-v2 read-back never inspects. Running after
+-- the outer user.deleted row is durable, any inner emit failure surfaces as a normal
+-- retryable processing_error on its own event row. (architect F1, migration 20260622212441)
+--
+-- Trigger DDL (in the migration, not this ref file):
+--   CREATE TRIGGER emit_grant_revocations_on_user_deleted_trigger
+--     AFTER INSERT ON public.domain_events FOR EACH ROW
+--     WHEN (NEW.event_type = 'user.deleted')
+--     EXECUTE FUNCTION public.emit_grant_revocations_on_user_deleted();
+CREATE OR REPLACE FUNCTION public.emit_grant_revocations_on_user_deleted()
+ RETURNS trigger
+ LANGUAGE plpgsql
+ SET search_path TO 'public', 'extensions', 'pg_temp'
+AS $function$
+DECLARE
+    v_grant record;
+BEGIN
+    FOR v_grant IN
+        SELECT id
+        FROM public.cross_tenant_access_grants_projection
+        WHERE consultant_user_id = NEW.stream_id
+          AND status = 'active'
+    LOOP
+        -- Handler process_access_grant_event keys its UPDATE on event_data->>'grant_id'
+        -- (NOT stream_id); set both. revoked_by NULL = automated (AsyncAPI contract
+        -- AccessGrantRevokedData.revoked_by made nullable in the same PR).
+        PERFORM api.emit_domain_event(
+            p_stream_id      := v_grant.id,
+            p_stream_type    := 'access_grant',
+            p_event_type     := 'access_grant.revoked',
+            p_event_data     := jsonb_build_object(
+                'grant_id',           v_grant.id,
+                'revoked_by',         NULL,
+                'revocation_reason',  'consultant_user_deleted',
+                'revocation_details', 'Consultant user was soft-deleted; grant auto-revoked by handle_user_deleted cascade.'
+            ),
+            p_event_metadata := jsonb_build_object(
+                'automated', true,
+                'source',    'handle_user_deleted_cascade',
+                'user_id',   NULL,
+                'reason',    'consultant_user_deleted'
+            )
+        );
+    END LOOP;
+
+    RETURN NULL;  -- AFTER trigger: return value ignored
+END;
+$function$;
--- a/infrastructure/supabase/handlers/trigger/emit_grant_revocations_on_user_deleted.sql
+++ b/infrastructure/supabase/handlers/trigger/emit_grant_revocations_on_user_deleted.sql
@@ -20,6 +20,10 @@ AS $function$
 DECLARE
     v_grant record;
 BEGIN
+    -- INVARIANT: a deleted user can only be a grant *consultant*, never a grant
+    -- *target* (targets are orgs/clients, never users), so consultant_user_id =
+    -- NEW.stream_id is the COMPLETE revoke set. Revisit for Phase N if any
+    -- user-as-target grant type is ever introduced.
     FOR v_grant IN
         SELECT id
         FROM public.cross_tenant_access_grants_projection
@@ -28,7 +32,9 @@ BEGIN
     LOOP
         -- Handler process_access_grant_event keys its UPDATE on event_data->>'grant_id'
         -- (NOT stream_id); set both. revoked_by NULL = automated (AsyncAPI contract
-        -- AccessGrantRevokedData.revoked_by made nullable in the same PR).
+        -- AccessGrantRevokedData.revoked_by made nullable in the same PR). READ-SIDE
+        -- POSTCONDITION: rows revoked via this path carry revoked_by IS NULL —
+        -- consumers must treat revoked_by as nullable.
         PERFORM api.emit_domain_event(
             p_stream_id      := v_grant.id,
             p_stream_type    := 'access_grant',
--- a/infrastructure/supabase/supabase/migrations/20260622212441_handle_user_deleted_cascade_cleanup_projections.sql
+++ b/infrastructure/supabase/supabase/migrations/20260622212441_handle_user_deleted_cascade_cleanup_projections.sql
@@ -1,0 +1,226 @@
+-- =============================================================================
+-- handle_user_deleted: cascade-cleanup FK-dependent membership/identity projections
+-- =============================================================================
+--
+-- Card: dev/active/handle-user-deleted-cascade-cleanup-projections/
+-- Architect verdict (software-architect-dbc, 2026-06-22): APPROVE WITH IN-PR FIXES.
+--   Decision record: ~/.claude/plans/fizzy-jingling-puppy-agent-ad12fa238108e55da.md
+--
+-- PROBLEM: handle_user_deleted only tombstones the user (deleted_at/is_active);
+-- it leaves stale rows in membership/identity projections, so soft-deleted users
+-- still resolve as members (e.g. check_user_org_membership) and consumers must add
+-- ad-hoc deleted_at IS NULL filters. Dev today: 9 stale rows across 2 users.
+--
+-- FIX (two layers, per architect F1):
+--   1. BEFORE handler handle_user_deleted: hard-DELETE the 7 membership/identity
+--      projections + NULL the contacts user-linkage. Order is LOAD-BEARING — the
+--      tombstone UPDATE runs FIRST so the recompute_user_accessible_organizations
+--      that fires on user_organizations_projection DELETE self-guards
+--      (WHERE deleted_at IS NULL) to a clean no-op. Audit-reference columns
+--      (*_by/created_by/...) are preserved everywhere.
+--   2. Grant-revoke is an AFTER-INSERT trigger (NOT a nested emit from the BEFORE
+--      handler): a nested emit_domain_event buries an inner failure in the INNER
+--      event's processing_error, which api.delete_user's outer Pattern-A-v2
+--      read-back never inspects -> a failed revoke would return {success:true}
+--      silently. The AFTER-INSERT trigger emits access_grant.revoked per active
+--      grant; any failure surfaces as a normal retryable processing_error.
+--
+-- Pitfall #6: deployed handle_user_deleted body (COALESCE tombstone + IF NOT FOUND
+-- RAISE P0002) is preserved verbatim; the cascade is appended. Pitfall #4 N/A
+-- (no INSERTs here). Pitfall #8 column-existence assertion included (Section 3).
+-- Idempotent/replay-safe: DELETE/NULL are no-ops on re-processing.
+-- =============================================================================
+
+
+-- -----------------------------------------------------------------------------
+-- Section 1 — handle_user_deleted: tombstone (preserved) + cascade DELETEs.
+-- -----------------------------------------------------------------------------
+CREATE OR REPLACE FUNCTION public.handle_user_deleted(p_event record)
+ RETURNS void
+ LANGUAGE plpgsql
+ SET search_path TO 'public', 'extensions', 'pg_temp'
+AS $function$
+BEGIN
+    -- TOMBSTONE FIRST (load-bearing: must precede the projection DELETEs so the
+    -- accessible_organizations recompute triggered by the user_organizations_
+    -- projection DELETE self-guards on deleted_at IS NULL to a no-op).
+    -- COALESCE order: existing tombstone wins (replay-safe), then event payload's
+    -- deleted_at, then event creation time as final fallback.
+    UPDATE public.users
+       SET deleted_at = COALESCE(
+             deleted_at,
+             (p_event.event_data->>'deleted_at')::timestamptz,
+             p_event.created_at
+           ),
+           is_active = false,
+           updated_at = p_event.created_at
+     WHERE id = p_event.stream_id;
+
+    IF NOT FOUND THEN
+        RAISE EXCEPTION 'User not found' USING ERRCODE = 'P0002';
+    END IF;
+
+    -- CASCADE: hard-DELETE membership/identity projections anchored on user_id.
+    -- These are derived "membership state" facts that no longer hold once the
+    -- user is tombstoned; reversibility on undelete is via re-onboarding. The
+    -- domain_events stream remains the audit trail. Audit-reference columns
+    -- (assigned_by, created_by, granted_by, ...) live on OTHER rows and are
+    -- untouched. NOTE: grant-revoke is handled by the AFTER-INSERT trigger in
+    -- Section 2, NOT here (no emit_domain_event in this synchronous body).
+    DELETE FROM public.user_roles_projection                    WHERE user_id = p_event.stream_id;
+    DELETE FROM public.user_organizations_projection            WHERE user_id = p_event.stream_id;
+    DELETE FROM public.user_notification_preferences_projection WHERE user_id = p_event.stream_id;
+    DELETE FROM public.user_addresses                           WHERE user_id = p_event.stream_id;
+    DELETE FROM public.user_phones                              WHERE user_id = p_event.stream_id;
+    DELETE FROM public.user_client_assignments_projection       WHERE user_id = p_event.stream_id;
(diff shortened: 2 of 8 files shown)


infrastructure/supabase/supabase/migrations/20260622212441_handle_user_deleted_cascade_cleanup_projections.sql
```diff
@@ -0,0 +1,226 @@
...
+    DELETE FROM public.user_organizations_projection            WHERE user_id = p_event.stream_id;
+    DELETE FROM public.user_notification_preferences_projection WHERE user_id = p_event.stream_id;
+    DELETE FROM public.user_addresses                           WHERE user_id = p_event.stream_id;
+    DELETE FROM public.user_phones                              WHERE user_id = p_event.stream_id;
+    DELETE FROM public.user_client_assignments_projection       WHERE user_id = p_event.stream_id;
+    DELETE FROM public.schedule_user_assignments_projection     WHERE user_id = p_event.stream_id;
+
+    -- contacts_projection: preserve the contact entity (it may be referenced by
+    -- other domain entities), drop only the user linkage. user_id is nullable;
+    -- the partial UNIQUE index idx_contacts_unique_user_id WHERE user_id IS NOT
```
lars-tice: **Nit (doc accuracy / pitfall-#8-adjacent):** This comment cites `idx_contacts_unique_user_id WHERE user_id IS NOT NULL` as the tolerating index. That index name does not exist. The actual partial indexes on `contacts_projection` are `idx_contacts_unique_user_per_org` UNIQUE on `(organization_id, user_id) WHERE user_id IS NOT NULL AND deleted_at IS NULL` and `idx_contacts_user_id` (non-unique). The NULL-link is still correct (both indexes are partial on `user_id IS NOT NULL`, so NULLing is tolerated), but the cited index name is phantom. **Fix:** correct the comment to `idx_contacts_unique_user_per_org (organization_id, user_id) WHERE user_id IS NOT NULL`. A future reader trusting this name could be misled. Same comment is duplicated in the reference file `handle_user_deleted.sql`.

infrastructure/supabase/supabase/migrations/20260622212441_handle_user_deleted_cascade_cleanup_projections.sql
```diff
@@ -0,0 +1,226 @@
...
+    -- NULL tolerates NULLs.
+    UPDATE public.contacts_projection SET user_id = NULL WHERE user_id = p_event.stream_id;
```
lars-tice: **Should-fix (replay completeness / soft-delete consistency):** `UPDATE contacts_projection SET user_id = NULL WHERE user_id = p_event.stream_id` does not filter `deleted_at IS NULL`. The partial UNIQUE index `idx_contacts_unique_user_per_org` is scoped `WHERE user_id IS NOT NULL AND deleted_at IS NULL`, so NULLing a contact row that is itself already soft-deleted is harmless but pointless churn. More importantly, the DELETE arms above hard-delete from `*_projection` tables while contacts is NULL-linked — confirm the asymmetry is intentional (contact entity preserved for cross-references; membership facts discarded). The comment justifies it; just flag that on a 2nd `user.deleted` replay this UPDATE re-touches 0 rows (idempotent) — fine. No functional bug; consider whether you want `AND user_id IS NOT NULL` to make the no-op explicit and avoid a write on replay.

infrastructure/supabase/supabase/migrations/20260622212441_handle_user_deleted_cascade_cleanup_projections.sql
```diff
@@ -0,0 +1,226 @@
...
+        -- Handler process_access_grant_event keys its UPDATE on event_data->>'grant_id'
+        -- (NOT stream_id); set both. revoked_by NULL = automated (contract made
+        -- nullable in this PR's AsyncAPI change).
+        PERFORM api.emit_domain_event(
+            p_stream_id      := v_grant.id,
+            p_stream_type    := 'access_grant',
+            p_event_type     := 'access_grant.revoked',
+            p_event_data     := jsonb_build_object(
+                'grant_id',           v_grant.id,
+                'revoked_by',         NULL,
```
lars-tice, line 129: **Nit (Design-by-Contract / HIPAA least-authority):** `revoked_by => NULL` for automated revocation is correct and now contract-valid (F2). Postcondition to document: the resulting `cross_tenant_access_grants_projection` row will have `revoked_by IS NULL` AND `revocation_reason='consultant_user_deleted'` — downstream consumers/queries that previously assumed `revoked_by` non-null (pre-F2 it was `required`) must tolerate NULL. The contract change covers the schema, but verify no `api.*` read RPC or frontend type still treats `revoked_by` as non-optional at the read boundary (the write contract is fixed; the read side is the latent risk). Recommend a grep of read RPCs returning grant rows.

infrastructure/supabase/supabase/migrations/20260622212441_handle_user_deleted_cascade_cleanup_projections.sql
```diff
@@ -0,0 +1,226 @@
...
+
+    RETURN NULL;  -- AFTER trigger: return value ignored
+END;
+$function$;
+
+DROP TRIGGER IF EXISTS emit_grant_revocations_on_user_deleted_trigger ON public.domain_events;
+CREATE TRIGGER emit_grant_revocations_on_user_deleted_trigger
+    AFTER INSERT ON public.domain_events
+    FOR EACH ROW
+    WHEN (NEW.event_type = 'user.deleted')
```
lars-tice, line 150: **Verified PASS — Design-by-Contract note for the trigger.** Precondition: a `user.deleted` row was just INSERTed (durable). Invariant: fires exactly once per `user.deleted` INSERT (WHEN clause is INSERT-only; dispatcher is BEFORE INSERT OR UPDATE but this trigger excludes UPDATE → no double-emit on the dispatcher's UPDATE of `processed_at`). Postcondition: one `access_grant.revoked` event emitted per active grant where `consultant_user_id = NEW.stream_id`; each is independently retryable via its own `processing_error`. This correctly matches the sanctioned AFTER-INSERT carve-out (`bootstrap_workflow_trigger`, `enqueue_workflow_from_bootstrap_event_trigger`) and does NOT violate the single-dispatcher rule. No change requested.

infrastructure/supabase/supabase/migrations/20260622212441_handle_user_deleted_cascade_cleanup_projections.sql
```diff
@@ -0,0 +1,226 @@
...
+-- event already exists; re-emitting would add audit noise). Covers all 3 arms.
+-- Migration context (not a trigger) -> emit_domain_event for grant-revoke is
+-- safe here. Dev targets today: 9 membership rows, 0 contacts, 0 active grants.
+-- -----------------------------------------------------------------------------
+DO $$
+DECLARE
+    v_user record;
+    v_grant record;
+BEGIN
+    FOR v_user IN SELECT id FROM public.users WHERE deleted_at IS NOT NULL LOOP
```
lars-tice, line 225: **Should-fix (backfill scope / blast radius):** The backfill iterates **every** `users WHERE deleted_at IS NOT NULL` and runs 7 DELETEs + 1 UPDATE per user. The header says "Dev targets today: 9 membership rows, 0 contacts, 0 active grants," but this loop is unbounded by row count and will run against whatever the target env holds at deploy time (prod could differ). DELETE/NULL are idempotent so re-running is safe, and the grant-revoke arm filters `status='active'` (so a 2nd run emits 0 duplicate `access_grant.revoked`) — good. **Ask:** (1) confirm this was validated against the *actual prod* deleted-user count, not just dev's 9; (2) the grant-revoke emit here runs in migration context (correct — `emit_domain_event` is safe outside the trigger), but each emit synchronously drives `process_access_grant_event` + `sync_accessible_organizations_from_grants` recompute inside the migration transaction — for a large prod backfill with many active grants this is a long single-txn. Low risk given 0 active grants on dev, but size it before prod.

infrastructure/supabase/supabase/migrations/20260622212441_handle_user_deleted_cascade_cleanup_projections.sql
```diff
@@ -0,0 +1,226 @@
...
+    END IF;
+
+    -- CASCADE: hard-DELETE membership/identity projections anchored on user_id.
+    -- These are derived "membership state" facts that no longer hold once the
+    -- user is tombstoned; reversibility on undelete is via re-onboarding. The
+    -- domain_events stream remains the audit trail. Audit-reference columns
+    -- (assigned_by, created_by, granted_by, ...) live on OTHER rows and are
+    -- untouched. NOTE: grant-revoke is handled by the AFTER-INSERT trigger in
+    -- Section 2, NOT here (no emit_domain_event in this synchronous body).
+    DELETE FROM public.user_roles_projection                    WHERE user_id = p_event.stream_id;
```
lars-tice, line 70: **Verified PASS — cascade completeness.** The 7 hard-DELETE targets (`user_roles_projection`, `user_organizations_projection`, `user_notification_preferences_projection`, `user_addresses`, `user_phones`, `user_client_assignments_projection`, `schedule_user_assignments_projection`) + contacts NULL-link are the correct membership/identity set, and all 8 tables + their `user_id` columns exist (baseline_v4 / schedule_template_refactor). Audit-reference columns (`assigned_by`, `granted_by`, etc.) live on OTHER entities' rows and are untouched — postcondition preserved. One FK to be aware of: `user_notification_preferences_projection.sms_phone_id REFERENCES user_phones ON DELETE SET NULL` — but since notification prefs are deleted in the same handler, ordering vs `user_phones` DELETE is moot. No over-delete. No change requested.

infrastructure/supabase/handlers/trigger/emit_grant_revocations_on_user_deleted.sql
```diff
@@ -0,0 +1,53 @@
...
+ SET search_path TO 'public', 'extensions', 'pg_temp'
+AS $function$
+DECLARE
+    v_grant record;
+BEGIN
+    FOR v_grant IN
+        SELECT id
+        FROM public.cross_tenant_access_grants_projection
+        WHERE consultant_user_id = NEW.stream_id
+          AND status = 'active'
```
lars-tice, line 31: **Nit (consultant-vs-target asymmetry — confirm intent):** This revokes only grants where the deleted user is the **consultant** (`consultant_user_id = NEW.stream_id`). If the deleted user could ever be a grant **target** (e.g. a client-side identity referenced by a grant), those grants are NOT revoked here. Per the current model (grants address consultants; targets are orgs/clients, not users) this is correct, but add a one-line comment stating the invariant: "deleted user is only ever a grant *consultant*; grant *targets* are orgs/clients, never users — so consultant_user_id is the complete revoke set." Documents the boundary for the Phase N (court/agency/family) grant types that may introduce user-targeted grants.

lars-tice (commented):
## Architectural Review — PR #81 (handle_user_deleted cascade cleanup)

**Verdict: APPROVE WITH IN-PR FIXES** (fixes are Nit/Should-fix only; no Blocking findings).

Independent re-review against the actual implementation (not the PR-body claims). The two-layer design is sound and the load-bearing claims verify out:

**Verified correct:**
- **Pitfall #6 (verbatim preservation):** the pre-PR `handle_user_deleted` body (COALESCE tombstone + `IF NOT FOUND RAISE P0002`) is byte-identical on `main`; the cascade is purely appended. ✅
- **Tombstone-first ordering (the crux):** `recompute_user_accessible_organizations` ends with `UPDATE public.users ... WHERE id = p_user_id AND deleted_at IS NULL`. Because the tombstone UPDATE runs first, the recompute fired by the `user_organizations_projection` DELETE (via `trg_sync_accessible_orgs` → `sync_accessible_organizations`) is a guaranteed no-op on the deleted user's own row. Invariant holds. ✅
- **AFTER-INSERT trigger semantics:** `AFTER INSERT ... FOR EACH ROW WHEN (NEW.event_type='user.deleted')` matches the sanctioned carve-out exactly (`bootstrap_workflow_trigger`, `enqueue_workflow_from_bootstrap_event_trigger` are all AFTER INSERT-only). The dispatcher `process_domain_event_trigger` is `BEFORE INSERT OR UPDATE`, but this new trigger is INSERT-only → **no double-emit on UPDATE.** Not a per-event-type-dispatcher-handler violation; it's a side-effect emitter. ✅
- **F1 retryable-processing_error rationale:** dispatcher `WHEN OTHERS` writes `processing_error`/`processing_error_detail` and never RAISEs out (Pattern A v2). Keeping the revoke OUT of the BEFORE handler and emitting one `access_grant.revoked` per grant AFTER the durable `user.deleted` row means each revoke is its own retryable event row, not swallowed by `api.delete_user`'s outer read-back. ✅
- **F2 nullable revoked_by:** `process_access_grant_event` revoked arm reads `safe_jsonb_extract_uuid(event_data,'revoked_by')` = `COALESCE((data->>key)::uuid, NULL)` → NULL-safe. `status='revoked'` is a valid CHECK value (baseline_v4 grants table CHECK = active/revoked/expired/suspended; the 4-value `terminated` CHECK at phase_2:86 belongs to `var_partnerships_projection`, not this table). PR #74 already aligned `revocation_reason`/`revocation_details`. ✅
- **AsyncAPI wiring:** `rbac.yaml` is the wired source (`asyncapi.yaml:159 $ref ./domains/rbac.yaml#/.../AccessGrantRevoked`); the PR correctly edited it. `generated-events.ts` `'revoked_by'?: string` matches `nullable:true` + removal from `required`. ✅
- **OTHER users' accessible_organizations on revoke:** the grant UPDATE (status→revoked) fires `sync_accessible_organizations_from_grants`, which enumerates affected users via OLD/NEW and recomputes each — correctly shrinking grant-sourced orgs for consultants targeted by the revoked grant. ✅
- **Pitfall #8:** fail-loud DO-block enumerates all 8 cascade columns + the 2 grant-trigger columns; matches the columns actually touched. ✅
- **Backfill/handler drift:** the 8-table cascade set is byte-identical between handler and backfill. Backfill grant-revoke filters `status='active'` → idempotent on re-run. ✅
- **Pitfall #4:** N/A confirmed (no INSERTs into projections; DELETE/UPDATE only).

**Items to address (in-PR):** see inline comments — all Nit/Should-fix.

Full console review echoed to the requesting session.

infrastructure/supabase/supabase/migrations/20260622212441_handle_user_deleted_cascade_cleanup_projections.sql
lars-tice: ✅ Fixed in de17f281 — corrected the comment to the real indexes `idx_contacts_unique_user_per_org` (organization_id, user_id) + `idx_contacts_user_id`, in both the migration and the `handle_user_deleted.sql` reference file.

infrastructure/supabase/supabase/migrations/20260622212441_handle_user_deleted_cascade_cleanup_projections.sql
lars-tice: ✅ Fixed in de17f281 — added the explicit `AND user_id IS NOT NULL` conjunct to the contacts UPDATE (handler + backfill + reference file), mirroring the partial-index predicates so replay is a guaranteed no-op.

infrastructure/supabase/supabase/migrations/20260622212441_handle_user_deleted_cascade_cleanup_projections.sql
lars-tice: ✅ Addressed in de17f281 — documented the read-side postcondition (rows revoked via this path carry `revoked_by IS NULL`; consumers must treat it as nullable) in the trigger. This also surfaced a real gap: the git-**tracked** `workflows/src/shared/types/generated/events.ts` was left stale (`AccessGrantRevokedData.revoked_by` still non-optional). Regenerated via `npm run sync-schemas`; workflows typecheck clean.

infrastructure/supabase/supabase/migrations/20260622212441_handle_user_deleted_cascade_cleanup_projections.sql
lars-tice: Thanks — DbC note captured. Locked the documented precondition in-code by adding the consultant-vs-target invariant comment to the trigger loop (de17f281).

infrastructure/supabase/supabase/migrations/20260622212441_handle_user_deleted_cascade_cleanup_projections.sql
lars-tice: ✅ Addressed in de17f281 — documented the env-dependent scope + idempotency inline, and added a `RAISE NOTICE` reporting the actual tombstoned-user / active-grant counts so the blast radius is observable in each environment's deploy log (rather than relying on dev's 9-row sample). The prod-count confirmation remains an operational pre-merge check.

infrastructure/supabase/supabase/migrations/20260622212441_handle_user_deleted_cascade_cleanup_projections.sql
lars-tice: Acknowledged — no change needed; cascade set confirmed complete.

infrastructure/supabase/handlers/trigger/emit_grant_revocations_on_user_deleted.sql
lars-tice: ✅ Fixed in de17f281 — added the invariant comment (a deleted user is only ever a grant *consultant*, never a *target*, so `consultant_user_id` is the complete revoke set) to both the trigger function and the reference file, flagged as a Phase-N boundary to revisit if a user-as-target grant type is ever introduced.

PR #86, opened 2026-07-01
docs(invitations): drop stale scalar `role` refs; consolidate on roles[] jsonb

lars-tice:
## Why

`invitations_projection.role` (scalar text) was dropped **2026-05-08** by migration `20260508170054` (superseded by the `roles` jsonb array). Deployed schema, all code paths, and generated types are already consolidated on `roles[]` — but the **table-reference doc had not caught up**. Uncovered during PR #85 invite-user routing UAT.

A scalar `role` is structurally wrong for this domain (invitations carry **0..N** roles; AsyncAPI `UserInvitedData` sets `roles: array, minItems: 0`). Its `role: null` reads previously caused the accept-page *"Invalid Invitation"* bug — the very reason it was dropped.

## What (docs-only — no code/schema change)

`documentation/infrastructure/reference/database/tables/invitations_projection.md`:
- Remove the `#### role (DEPRECATED)` section → replace with an audit note (dropped 2026-05-08, why a scalar was wrong).
- Fix the false invariant *"one of `role` or `roles` should be populated"* → `roles` defaults to `'[]'::jsonb`; empty array valid (matches AsyncAPI `minItems: 0`).
- Fix **4** SQL example `SELECT`s + the `handle_user_invited` INSERT excerpt that referenced the dropped column (would now error).
- Bump `last_updated`.

## Verification

- `grep` for a scalar `role` column / `->>'role'` extraction / `#### role` header → none.
- Every SQL snippet now references only existing columns.
- `npm run docs:check` → passes.

Out of scope (intentionally): `baseline_v4.sql` still lists `role` — it's a frozen Day-0 snapshot; the later incremental migration is the source of truth.

### DIFF
--- a/documentation/infrastructure/reference/database/tables/invitations_projection.md
+++ b/documentation/infrastructure/reference/database/tables/invitations_projection.md
@@ -1,6 +1,6 @@
 ---
 status: current
-last_updated: 2026-02-19
+last_updated: 2026-07-01
 ---
 
 <!-- TL;DR-START -->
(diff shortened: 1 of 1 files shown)


lars-tice (commented):
As expected

PR #87, opened 2026-07-01
docs(frontend): command-feedback standard — banner-authoritative, toast-as-visual-echo (Phase 1)

lars-tice:
## Why (Phase 1 of the frontend command-feedback codification)

Revisits and supersedes the `codify-frontend-feedback-and-testing-standards` seed card's S1, whose premise was **inverted by the codebase**. A full audit (2026-07-01) found banners are the house style (~35 sites/15 files), toasts a 3-page minority — and `UsersManagePage` runs **three conflicting failure policies** including **three unmanaged screen-reader double-announce bugs**.

`software-architect-dbc` reviewed the reframed decision with the full evidence map and the user's proposed model.

## The standard: "Banner-authoritative, toast-as-visual-echo"

Endorses the user's `success = banner` / `failure = toast + banner` model, with the a11y refinement that resolves the double-announce:

| Outcome | Banner | Toast |
|---|---|---|
| **Success** | `role="status"` (polite), persistent — **banner only** | — |
| **Failure** | `role="alert"` (assertive), persistent — **owns the one announcement** | **`aria-hidden`**, `duration:Infinity` — **visual scroll-independence only** |

**Load-bearing fact** (verified, version-pinned): the installed **Sonner v2 announces politely** (container `aria-live="polite"`), never assertively — so the seed card's rule #4 ("toast is `alert`, demote the banner") was backwards. Banner-authoritative keeps the assertive interrupt errors deserve and avoids stripping `role="alert"` from ~35 existing banners.

## What's in this PR (docs-only)

- **New standard**: `documentation/frontend/patterns/command-feedback.md` — DbC a11y contract (INV-1 single announcement / INV-2 no orphan speech / INV-3 durability), decision tree (banner vs toast vs reserved modal), display-layer error sanitization, focus + reduced-motion, mandated `data-testid`s, Adoption-status block.
- **Discoverability** (per AGENT-GUIDELINES): `AGENT-INDEX.md` task-tree row + keyword rows + Frontend catalog entry; `frontend/CLAUDE.md` accessibility forward-link.
- **Seed cards**: corrected S1 rule #4 (struck the inverted premise) + "Revisited" banner; annotated the duplicate toast/banner UAT card **SUPERSEDED**.

## Satisfies the three hard constraints
1. **All errors surfaced** — every failure reaches the assertive banner (+ visual toast).
2. **Discoverable/documented** per AGENT-GUIDELINES — frontmatter, TL;DR, AGENT-INDEX, CLAUDE.md link.
[...]

PR #88, opened 2026-07-02
feat(frontend): command-feedback Phase 2 — abstraction + UsersManagePage reference impl

lars-tice:
## Why

Phase 2 of the command-feedback standard (doc shipped in #87). Builds the shared abstraction and migrates `UsersManagePage` as the reference implementation, fixing the concrete a11y defects the audit found: **3 unmanaged screen-reader double-announce sites**, the notif-prefs `clearError()` workaround, and the invite success/failure asymmetry.

Standard: [`documentation/frontend/patterns/command-feedback.md`](../blob/main/documentation/frontend/patterns/command-feedback.md).

## What (4 commits, each independently green)

**Abstraction (additive)**
- `sanitizeCommandError(raw, fallback?)` — pure display-layer guard (masks `Event processing failed:` / `ERRCODE` / SQLSTATE codes → friendly fallback; preserves `raw` for `log.warn`). **8 unit tests.**
- `<CommandFeedbackBanner kind="error|success">` — generic authoritative banner (error→`role="alert"`, success→`role="status"`), `forwardRef` + `tabIndex=-1` for focus-to-banner, mandated `data-testid`s.
- `useCommandFeedback()` — orchestrates presentation only (no VM bypass): sanitizes + `log.warn`s, fires the persistent **`aria-hidden`** failure toast echo for scroll-independence. **INV-2**: renders the echo with `closeButton:false` + a scoped `cf-echo-toast` class, and a once-installed, class-targeted `MutationObserver` sets `aria-hidden="true"` (also silences the polite Sonner container → INV-1) and `tabIndex="-1"` on **our echo toasts only**.

**Reference migration — `UsersManagePage`**
- **Failure**: dropped the redundant `toast.error` that announced alongside the `role="alert"` banner (the double-announce). Now sanitized banner (announces once) + silent echo. Also `clearError()`s the VM's raw error first so the sanitized message wins over `viewModel.error` on the banner.
- **Success**: invite / deactivate / reactivate / delete / notif-prefs now use a polite `role="status"` banner instead of a toast (success = banner per the standard). Invite preserves its `invite-success-${action}` test id.
- Removed the notif-prefs `clearError()` double-aria-live hack and the now-dead `sonner` import. **Zero `toast.*` remain on the page.**

## Verification
`tsc --noEmit` clean · `eslint` clean · `npm run build` passes · `sanitizeCommandError` 8/8.

## Acceptance gate (before merge)
- [ ] **`@axe-core/playwright`** on `UsersManagePage`: **single announcement** on failure + **zero `aria-hidden-focus`** — the run-verification of the INV-2 neutralizer (per the doc DoD).
- [ ] Manual NVDA/VoiceOver pass.

## Follow-on (this PR or a fast-follow)
- Invite **failure** path (`formViewModel.submissionError` banner) → add the `aria-hidden` echo + focus-to-banner (`useEffect`), and fold the hand-rolled banner into `<CommandFeedbackBanner>`.
- Phase 3: progressive rollout to the other ~14 command-feedback pages (banners already surface every failure, so partial rollout is never broken — only less uniform).

### DIFF
--- a/frontend/src/hooks/useCommandFeedback.ts
+++ b/frontend/src/hooks/useCommandFeedback.ts
@@ -1,0 +1,118 @@
+/**
+ * useCommandFeedback — orchestrates the *presentation* of command results per
+ * the command-feedback standard (documentation/frontend/patterns/command-feedback.md).
+ *
+ * Division of labor:
+ *   - The **banner** (`<CommandFeedbackBanner>`, driven by ViewModel state) is the
+ *     authoritative surface and owns the single ARIA announcement (INV-1).
+ *   - This hook adds the **failure toast echo** — an `aria-hidden`, non-announcing,
+ *     persistent (`duration: Infinity`) visual echo so a failure is seen even when
+ *     the banner is scrolled off-screen — and sanitizes + logs the raw error.
+ *   - It writes NOTHING to domain state; all mutations stay on the ViewModel path.
+ *
+ * INV-2 (no focusable content under `aria-hidden`): Sonner v2 renders each toast as
+ * `<li tabIndex=0>` plus (optionally) a focusable close button. We render the echo
+ * with `closeButton: false` and a scoped class, then a once-installed MutationObserver
+ * neutralizes our echo toasts only — setting `aria-hidden="true"` (also silences the
+ * polite container announcement, satisfying INV-1) and `tabIndex="-1"`.
+ *
+ * Success has NO toast (banner only). Focus-to-banner on form-blocking failures is
+ * the consumer's responsibility (a `useEffect` keyed on the error — never setTimeout).
+ */
+
+import { useCallback, useRef } from 'react';
+import { toast } from 'sonner';
+import { sanitizeCommandError } from '@/utils/sanitizeCommandError';
+import { Logger } from '@/utils/logger';
+
+const log = Logger.getLogger('component');
+
+/** Class applied to echo toasts so the neutralizer targets only them. */
+export const CF_ECHO_CLASS = 'cf-echo-toast';
+
+export interface CommandFailureOptions {
+  /** Operation-specific friendly fallback when the raw error looks internal. */
+  fallback?: string;
+  /** Correlation id for the log line (never displayed). */
+  correlationId?: string;
+}
+
+export interface UseCommandFeedbackResult {
+  /**
+   * Handle a command failure: sanitize + `log.warn` the raw error, fire the
+   * `aria-hidden` echo toast, and return the display string for the banner.
+   */
+  failed: (raw: unknown, opts?: CommandFailureOptions) => string;
+  /** Success path — no toast; dismisses any lingering echo. */
+  succeeded: () => void;
+  /** Dismiss the paired echo toast (call when the banner is dismissed/cleared). */
+  clear: () => void;
+}
+
+let neutralizerInstalled = false;
+
+function neutralize(li: Element): void {
+  li.setAttribute('aria-hidden', 'true');
+  (li as HTMLElement).tabIndex = -1;
+}
+
+function applyNeutralize(root: Element): void {
+  root.querySelectorAll?.(`.${CF_ECHO_CLASS}`).forEach((el) => {
+    const li =
+      el.closest('li[data-sonner-toast]') ?? (el.matches('li[data-sonner-toast]') ? el : null);
+    if (li) neutralize(li);
+  });
+}
+
+/** Install (once) a scoped observer that strips focusability + announcement from echo toasts. */
+function ensureEchoNeutralizer(): void {
+  if (neutralizerInstalled || typeof document === 'undefined') return;
+  neutralizerInstalled = true;
+  const observer = new MutationObserver((mutations) => {
+    for (const m of mutations) {
+      m.addedNodes.forEach((n) => {
+        if (n.nodeType !== 1) return;
+        applyNeutralize(n as Element);
+      });
+    }
+  });
+  observer.observe(document.body, { childList: true, subtree: true });
+  applyNeutralize(document.body); // catch any echo already mounted
+}
+
+/**
+ * @param scope Optional label prefixing log lines (e.g. 'users', 'roles').
+ */
+export function useCommandFeedback(scope = 'command'): UseCommandFeedbackResult {
+  const echoId = useRef<string | number | null>(null);
+
+  const dismissEcho = useCallback(() => {
+    if (echoId.current != null) {
+      toast.dismiss(echoId.current);
+      echoId.current = null;
+    }
+  }, []);
+
+  const failed = useCallback(
+    (raw: unknown, opts?: CommandFailureOptions): string => {
+      const { display, raw: rawStr } = sanitizeCommandError(raw, opts?.fallback);
+      log.warn(`${scope}: command failed`, {
+        raw: rawStr,
+        correlationId: opts?.correlationId,
+      });
+      dismissEcho();
+      ensureEchoNeutralizer();
+      echoId.current = toast.error(display, {
+        duration: Infinity,
+        closeButton: false,
+        className: CF_ECHO_CLASS,
+      });
+      return display;
+    },
+    [scope, dismissEcho]
+  );
+
+  const succeeded = useCallback(() => dismissEcho(), [dismissEcho]);
+
+  return { failed, succeeded, clear: dismissEcho };
+}
--- a/frontend/src/hooks/useCommandFeedback.ts
+++ b/frontend/src/hooks/useCommandFeedback.ts
@@ -5,30 +5,23 @@
  * Division of labor:
  *   - The **banner** (`<CommandFeedbackBanner>`, driven by ViewModel state) is the
  *     authoritative surface and owns the single ARIA announcement (INV-1).
- *   - This hook adds the **failure toast echo** — an `aria-hidden`, non-announcing,
- *     persistent (`duration: Infinity`) visual echo so a failure is seen even when
- *     the banner is scrolled off-screen — and sanitizes + logs the raw error.
+ *   - This hook sanitizes + logs the raw error and drives the **failure toast echo** —
+ *     an `aria-hidden`, non-announcing, persistent visual echo (rendered by the page via
+ *     `<CommandFeedbackEcho message={echoMessage} />`) so a failure is seen even when the
+ *     banner is scrolled off-screen.
  *   - It writes NOTHING to domain state; all mutations stay on the ViewModel path.
  *
- * INV-2 (no focusable content under `aria-hidden`): Sonner v2 renders each toast as
- * `<li tabIndex=0>` plus (optionally) a focusable close button. We render the echo
- * with `closeButton: false` and a scoped class, then a once-installed MutationObserver
- * neutralizes our echo toasts only — setting `aria-hidden="true"` (also silences the
- * polite container announcement, satisfying INV-1) and `tabIndex="-1"`.
- *
- * Success has NO toast (banner only). Focus-to-banner on form-blocking failures is
- * the consumer's responsibility (a `useEffect` keyed on the error — never setTimeout).
+ * The echo is a plain, non-Sonner `aria-hidden` element, so INV-2 (no focusable content
+ * under `aria-hidden`) holds by construction — there is nothing to neutralize. Success has
+ * NO echo (banner only). Focus-to-banner on form-blocking failures is the consumer's
+ * responsibility (a `useEffect` keyed on the error — never setTimeout).
  */
 
-import { useCallback, useRef } from 'react';
-import { toast } from 'sonner';
+import { useCallback, useState } from 'react';
 import { sanitizeCommandError } from '@/utils/sanitizeCommandError';
 import { Logger } from '@/utils/logger';
 
 const log = Logger.getLogger('component');
-
-/** Class applied to echo toasts so the neutralizer targets only them. */
-export const CF_ECHO_CLASS = 'cf-echo-toast';
 
 export interface CommandFailureOptions {
   /** Operation-specific friendly fallback when the raw error looks internal. */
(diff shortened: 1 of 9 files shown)


frontend/src/hooks/useCommandFeedback.ts
```diff
@@ -0,0 +1,118 @@
...
+
+let neutralizerInstalled = false;
+
+function neutralize(li: Element): void {
+  li.setAttribute('aria-hidden', 'true');
+  (li as HTMLElement).tabIndex = -1;
+}
+
+function applyNeutralize(root: Element): void {
+  root.querySelectorAll?.(`.${CF_ECHO_CLASS}`).forEach((el) => {
```
lars-tice: 🔴 **F1 — BLOCKER.** `root.querySelectorAll('.cf-echo-toast')` matches **descendants only, never the root element**. Sonner v2.0.7 puts the per-toast `className` on the `<li data-sonner-toast>` itself, and the observer (`:75`) passes that `<li>` as the *added node* — so `applyNeutralize(li)` searches the li's children, never finds the class, and `neutralize()` never runs. Empirically reproduced (jsdom): added-node-is-the-li → `tabIndex 0, aria-hidden null` (BROKEN). Result: the echo keeps `tabIndex=0` (`aria-hidden-focus` / WCAG 4.1.2) **and** is announced by Sonner's polite container → the double-announce this PR kills is reintroduced.

Minimal fix — check the root itself:
```ts
function applyNeutralize(root: Element): void {
  const hit = (el: Element) => {
    const li = el.closest('li[data-sonner-toast]');
    if (li) neutralize(li);
  };
  if (root.matches?.(`.${CF_ECHO_CLASS}`)) hit(root);
  root.querySelectorAll?.(`.${CF_ECHO_CLASS}`).forEach(hit);
}
```
**Recommended instead:** drop Sonner for the echo (F2) — a plain fixed-position `<div aria-hidden="true" data-testid="command-feedback-toast-error">` fixes F1+F2+F3 and removes this whole observer. Either way, the axe gate (F4) must actually run before merge.

frontend/src/hooks/useCommandFeedback.ts
```diff
@@ -0,0 +1,118 @@
...
+  neutralizerInstalled = true;
+  const observer = new MutationObserver((mutations) => {
+    for (const m of mutations) {
+      m.addedNodes.forEach((n) => {
+        if (n.nodeType !== 1) return;
+        applyNeutralize(n as Element);
+      });
+    }
+  });
+  observer.observe(document.body, { childList: true, subtree: true });
```
lars-tice: 🟠 **F2 — HIGH.** `childList`-only observation never reacts to *attribute* mutations. Sonner re-renders each toast on hover/height-recalc/index-shift; persistence survives today only because Sonner passes a constant `tabIndex:0` (React skips re-committing) — an undocumented coupling to reconciliation internals that breaks silently if Sonner computes `tabIndex` conditionally or remounts the `<li>`. A module-global `MutationObserver` on `document.body{subtree:true}` for the whole app lifetime, to fight a library re-setting one attribute, is a lot of machinery. Prefer the non-Sonner `<div aria-hidden>` echo — a static element gives INV-2 for free.

frontend/src/hooks/useCommandFeedback.ts
```diff
@@ -0,0 +1,118 @@
...
+      log.warn(`${scope}: command failed`, {
+        raw: rawStr,
+        correlationId: opts?.correlationId,
+      });
+      dismissEcho();
+      ensureEchoNeutralizer();
+      echoId.current = toast.error(display, {
+        duration: Infinity,
+        closeButton: false,
+        className: CF_ECHO_CLASS,
```
lars-tice: 🟡 **F3 — MEDIUM.** The echo is created with only `className: CF_ECHO_CLASS` — the standard's mandated `data-testid="command-feedback-toast-error"` is absent, and `toast.error(...)` options can't pass arbitrary DOM attributes. The acceptance Playwright test (F4) needs a stable selector for "single announcement." A non-Sonner `<div>` echo (F2) carries the testid natively.

lars-tice (commented):
## 🏛️ Architectural Review — software-architect-dbc

**Verdict: REQUEST CHANGES.** The abstraction is well-structured and the migration is largely faithful, but the **load-bearing INV-2/INV-1 neutralizer is demonstrably non-functional as written** (empirically reproduced in jsdom against the exact logic). Since eliminating the double-announce + satisfying `aria-hidden-focus` is the entire point of this PR, this must be fixed and axe-proven before merge. `tsc`/`eslint`/`build` are green but say nothing about this runtime DOM-selector bug. Fix is small; happy to flip to APPROVE once F1 is corrected and the axe gate is green.

### 🔴 F1 — BLOCKER · the neutralizer never matches the echo `<li>` (`useCommandFeedback.ts:59-60`)
Sonner v2.0.7 places the per-toast `className` **directly on the `<li data-sonner-toast>`** (verified in `node_modules/sonner/dist/index.mjs`). The observer hands each **added node** to `applyNeutralize(n)`; when a toast mounts into the already-present `<ol>`, the added node **is that `<li>`**. But `applyNeutralize` runs `root.querySelectorAll('.cf-echo-toast')` — which matches **descendants only, never the root**. So the `<li>` carrying the class is never found; `aria-hidden` is never set and `tabIndex` stays `0`.

```
jsdom repro of the PR logic:
 CASE A (added node = the li):     tabIndex 0   aria-hidden null   ← BROKEN
 CASE B (added node = ancestor):   tabIndex -1  aria-hidden true   ← works
```
Sonner mounts a position's `<ol>` as soon as it has ≥1 toast, so only the *first* echo can incidentally hit CASE B; every subsequent echo hits CASE A. The once-only `applyNeutralize(document.body)` sweep runs before any toast mounts, so it never helps. **Consequence:** echo `<li>` keeps `tabIndex=0` (the exact `aria-hidden-focus` WCAG 4.1.2 defect — and it isn't even under `aria-hidden`) **and** Sonner's polite `aria-live` container announces it → the **double-announce this PR sets out to kill is reintroduced** (INV-1). *Remediation: match the root itself (`root.matches(...)` / `closest`), OR — recommended — adopt the non-Sonner echo (F2), which fixes F1+F2+F3 at once.*

### 🟠 F2 — HIGH · observer watches `childList` only; attribute re-assertion unguarded (`useCommandFeedback.ts:79`)
Even with F1 fixed, the observer never reacts to *attribute* mutations. Sonner re-renders toasts often (hover, height recalc, index shifts). Persistence survives today only because Sonner passes a constant `tabIndex:0` so React skips re-committing it — an undocumented dependency on reconciliation internals that breaks silently if Sonner ever computes `tabIndex` conditionally or remounts the `<li>`. **Recommended:** drop Sonner for the echo and render a plain fixed-position `<div aria-hidden="true">` (the doc explicitly sanctions "or render a non-Sonner echo element"). The echo has no interactive affordance, so it needs nothing Sonner provides but positioning — and this removes the whole MutationObserver apparatus, fixes F1, and gives a native testid (F3).

### 🟡 F3 — MEDIUM · mandated `command-feedback-toast-error` testid absent on the echo (`useCommandFeedback.ts:105-108`)
The standard's testid section + DoD require it, and the acceptance Playwright test needs it. `toast.error(...)` options can't pass arbitrary DOM attributes. Folding into F2's `<div>` echo resolves it.

### 🟡 F4 — MEDIUM (process) · the axe/AT acceptance gate is the only thing that would have caught F1
Merging on green `tsc`/`eslint`/`build` alone would ship a fully broken neutralizer. Per the doc DoD, `@axe-core/playwright` (single-announcement + zero `aria-hidden-focus`) + a manual NVDA/VoiceOver pass are **mandatory before merge** — blocking, not follow-on.

### 🟡 F5 — MEDIUM · invite/save `submissionError` bypasses the standard — raw leak + INV-1 gap (`UsersManagePage.tsx:~797-824`, `:719`)
The hand-rolled invite/save `role="alert"` banners render `formViewModel.submissionError` and `submissionErrorDetails.details` **verbatim** (a `font-mono` block) — the exact handler-internal leak `sanitizeCommandError` exists to stop. Separately, the success banner guard (`!operationError && !viewModel.error`) does **not** exclude `submissionError` — safe today only because `UserFormViewModel.submit` clears it at entry (incidental, not structural). *In-PR (address-inline convention): route `submissionError` through `sanitizeCommandError` and add `!formViewModel?.submissionError` to the success-banner guard.*

### 🟡 F6 — MEDIUM · "all success paths → banner" is inaccurate: edit-mode save is silent (`UsersManagePage.tsx:379-398`)
`handleSubmit` calls `showCommandSuccess` only when `panelMode === 'create'`. A successful **edit-mode profile save / role modification** shows no success banner (the stale comment at `UserFormViewModel.ts:1007` even claims something sets `successMessage = 'Roles updated'` — nothing does). Silent before this PR too (not a regression), but the commit claims every success uses a banner, and this is the *reference* page. Wire `showCommandSuccess('Changes saved')`/`'Roles updated'`, or document the omission.

### 🟢 F7 — LOW · sanitizer marker gaps (`sanitizeCommandError.ts:38`)
False negative: `\b[A-Z]\d{4}\b` catches `P9002` but not a bare standard SQLSTATE `23505`. False positive: masks legit messages with unit/room/order codes ("Suite B4021") — all-or-nothing. Low incidence; note it.

### 🟢 F8 — LOW · invite-success drops the mandated container testid (`UsersManagePage.tsx:720-723` + `CommandFeedbackBanner.tsx:79`)
`data-testid={successTestId}` overrides the default, so invite successes carry `invite-success-${action}` but lose `command-feedback-banner`. Consider emitting both.

### 🟢 F9 — LOW · doc drift (`command-feedback.md:90,112`)
This PR *is* Phase 2 landing, yet the doc still reads "Planned (Phase 2 — not yet implemented)" / "⏳ Phase 2." Flip to in-progress/implemented; reference the shipped hook/util/component. Reduced-motion clause for the echo also unaddressed (low).

### 🔵 F10 — INFO
`succeeded` is dead API surface (page uses `clear`). `CommandFeedbackBanner`'s `forwardRef`+`tabIndex=-1` focus affordance is presently inert (correct for non-blocking list-row failures; forward-looking for Phase 3 form-blocking).

---

### ✅ Verified correct (not assumed)
- **INV-1 banner-vs-banner exclusion** sound: helpers are strictly mutually exclusive; success guard `!operationError && !viewModel.error` is a hard backstop; role-violation variants set `viewModel.error` → suppress the success banner. ✓
- **Raw-masking fix is real & correct**: `UsersErrorBanner` renders `{error || operationError}` (VM error wins), so `clearError()` before `setOperationError` is necessary. ✓
- **`clearError()` clears role-violation state, but the role-modification path does NOT route through the new helpers** (it goes via `formViewModel.submit` → `viewModel.error`), so it's untouched and intact. ✓
- **`closeButton:false` wins** (no Toaster-level default in `App.tsx`) → one fewer focusable. ✓
- **Hooks/MobX**: `failed`/`succeeded`/`clear` referentially stable; helper deps correct; no stale closures; `observer()` intact. ✓
- **DbC/anti-pattern**: presentation-only (no VM bypass); a deliberately multi-caller abstraction (not the S2 one-caller anti-pattern); hook correctly not unit-tested (VM + pure sanitizer are). ✓ *Caveat: precisely because the neutralizer is untested here, F1 slipped — the argument for the axe gate.*

### Required before merge
1. **F1** — fix `applyNeutralize` (match root) or adopt the non-Sonner `<div>` echo (also clears F2/F3).
2. **F4** — run `@axe-core/playwright` (single-announcement + zero `aria-hidden-focus`) + manual AT. Non-negotiable given F1.
3. **F3** — echo `command-feedback-toast-error` testid.
4. **F5** — sanitize `submissionError` + tighten the success guard (in-PR).

**Recommended in-PR:** F6, F9. **Defer OK:** F7, F8, F10, reduced-motion.

<sub>🤖 software-architect-dbc review, teed inline + to console.</sub>

lars-tice:
## Review fixes applied — `517bb3d9`

Thanks — the BLOCKER was a real bug and the empirical repro nailed it. Pivoted per your F2 recommendation.

- **F1 / F2 / F3 — ✅ pivoted to a non-Sonner echo.** Deleted the `MutationObserver` + Sonner coupling entirely. `<CommandFeedbackEcho>` is a fixed-position, `aria-hidden`, non-focusable `<div>` rendered from hook state — **INV-2 holds by construction** (nothing focusable to neutralize, so F1's selector-bug class can't recur), and it carries `data-testid="command-feedback-toast-error"` natively (F3). `useCommandFeedback` now holds `echoMessage` state. **4 new component tests** assert `aria-hidden` + zero focusable descendants + renders-nothing-when-empty.
- **F5 — ✅** both hand-rolled invite/edit submission banners now render `sanitizeCommandError(submissionError).display`; the raw `font-mono submissionErrorDetails` leak is **removed**; success-banner guard tightened with `!formViewModel?.submissionError` (INV-1 now structural).
- **F6 — ✅** edit-mode save success now shows a `'Changes saved'` `role="status"` banner (was silent).
- **F9 / doc — ✅** `command-feedback.md` updated to the shipped non-Sonner echo (INV-2/INV-3/reduced-motion/dismissal reworded; the rejected Sonner approach recorded), Planned→Implemented, adoption status Phase 2 in-progress, `last_updated` bumped — per AGENT-GUIDELINES.
- **Deferred (per your defer-OK):** F7 (5-digit SQLSTATE — skipped to avoid the false-positive on legit 5-digit codes), F8 (dual testid), F10.

**Remaining merge gate (F4):** the mechanism is now proven *by construction* + the component test, but the integrated **`@axe-core/playwright`** single-announcement + `aria-hidden-focus` run and a manual NVDA/VoiceOver pass still stand as the acceptance gate (`@axe-core/playwright` isn't yet a dep).

Verified: `tsc` clean, `eslint` clean, `build` passes, **12/12 tests**.

lars-tice:
## 🏛️ Architect re-review @ `517bb3d9` — **APPROVE** (mergeable pending only the F4 axe/AT gate)

The Sonner pivot is clean and correct; resolving F1 *by construction* (rather than patching the fragile neutralizer) is the right call. Every prior finding is genuinely resolved in the current tree (not just the commit message), and the pivot introduced no regressions. `tsc` clean · `eslint` clean · 12/12 unit tests.

### Per-finding status
| # | Prior sev | Status |
|---|---|---|
| **F1** | BLOCKER | ✅ Resolved — echo is a plain `aria-hidden` `<div>`, text only; INV-2 holds by construction (nothing focusable to match/miss) |
| **F2** | HIGH | ✅ Moot — MutationObserver + `cf-echo-toast` + Sonner coupling fully deleted |
| **F3** | MED | ✅ `command-feedback-toast-error` native on the echo |
| **F5** | MED | ✅ **Both** banners (`:813` invite, `:1115` edit) now render `sanitizeCommandError(...).display`; raw `details` leak deleted from both; success guard tightened (`:724`) |
| **F6** | MED | ✅ `'Changes saved'` in the edit branch (`:402`); verified it **cannot** fire on partial-failure (`result.success` false → else skipped) |
| **F9** | LOW | ✅ Doc reworded to the non-Sonner echo; Planned→Implemented; adoption 🚧 Phase 2; `last_updated` bumped |
| F7 / F8 / F10 | defer-OK | ⏸️ Unchanged, fine to defer |
| **F4** | GATE | ⛔ Open — `@axe-core/playwright` single-announcement + `aria-hidden-focus` + manual AT. Running-app only; not dischargeable via static review. Mechanism now proven by construction + component test → low risk |

### Pivot checks (all clean)
Referential stability of `failed`/`clear`/`succeeded` ✓ · return-shape consumed correctly ✓ · MobX reactivity (observable `submissionError` read in an `observer` page; `echoMessage` is React state) ✓ · INV-3 + mutual exclusion (helpers clear the opposite + echo) ✓ · null handling ✓.

### New findings (all LOW / non-blocking)
- **N1** — the display-tier `sanitizeCommandError(...).display` in JSX drops `.raw`, but raw is already `log.warn`'d upstream at the VM tier — acceptable, no leak.
- **N2** (pre-existing) — `UsersViewModel.successMessage` is a write-only observable (rendered nowhere); stale comment at `UserFormViewModel.ts:1007` now misleads. Phase-3 cleanup candidate.
- **N3** (nit) — doc `:37` lead-in + page comment `:130` still say "toast" — wording polish.
- **N4** (pre-existing, not a regression) — the hand-rolled `submissionError` `role="alert"` banner and `UsersErrorBanner` could theoretically co-mount (narrow window; `selectAndLoadUser` clears `operationError`). Exactly the double-announce class this epic targets → good Phase-3 seed.

**Bottom line:** ship-ready pending the F4 axe-core + manual AT run. N1–N4 non-blocking; N2/N4 are Phase-3 seeds.

lars-tice:
## ✅ F4 gate discharged — verified on the live deployment (`a22bcab5`)

The merge auto-deployed via `frontend-deploy.yml` (Deploy Frontend → `a4c.firstovertheline.com`). Ran the command-feedback failure path there (deactivate forced to fail via DevTools "Go offline") and checked the invariants in the console — **all pass**:

| Check | Result |
|---|---|
| **INV-2** · focusable under `aria-hidden` (want 0) | **0** ✅ |
| echo present | **true** ✅ |
| echo `aria-hidden` | **"true"** ✅ |
| echo itself focusable (want false) | **false** ✅ |
| **INV-1** · visible `role="alert"` (want 1) | **1** ✅ |
| raw internal leak (want false) | **false** ✅ |

`echo_aria_hidden: "true"` is the programmatic equivalent of the Accessibility-tree "not exposed" check — the echo is out of the a11y tree (won't announce) while exactly one banner announces. This is the machine half of F4 (single-announcement + zero `aria-hidden-focus`) confirmed on the real build, no screen reader required. The manual audio AT pass remains an optional footnote (the DOM already dictates the outcome).

Verified without touching data — the offline-forced deactivate never reached the backend, so the fixture is unchanged.

PR #90, opened 2026-07-21
feat(command-feedback): Phase 3 PR1 — form-submit single-announce (N4) + successMessage cleanup (N2)

lars-tice:
## Context

First PR of the **command-feedback standard Phase 3 rollout** (`documentation/frontend/patterns/command-feedback.md`, shipped #87/#88/#89). Applies the already-reviewed contract to the two remaining gaps on the reference `UsersManagePage` before the sibling-page rollout (PR 2) and Tier-2 sweep (PR 3).

## N4 — form-submit routed through the single-announcement path
- Both hand-rolled invite/edit `submissionError` `role="alert"` banners → shared `<CommandFeedbackBanner kind="error">` (testids `invite-submission-error` / `edit-submission-error`), sanitized with an operation-specific fallback.
- `handleSubmit` failure branch clears any stale page-level `operationError` and fires the `aria-hidden` echo (+ `log.warn`) for scroll-independence.
- **INV-1 structural guard**: page-level `UsersErrorBanner` guarded with `!formViewModel?.submissionError` → at most one `role="alert"` region ever mounts. Rich role-violation / partial-failure variants are unaffected (they suppress `submissionError` by construction).
- **Form-blocking focus-to-banner** wired via `useEffect` keyed on `submissionError` → `CommandFeedbackBanner`'s `forwardRef`/`tabIndex={-1}` (never `setTimeout`).
- F4 a11y spec (`e2e/command-feedback-a11y.spec.ts`) invite-failure test extended: banner focused + echo fires (aria-hidden) + single alert, asserted via testids (the echo now shares banner text, so `getByText` would be ambiguous).

## N2 — remove write-only `UsersViewModel.successMessage`
- Deleted the observable + all 17 setters (0 readers; the page-level `showCommandSuccess` banner is the authoritative success surface) + the now-dead no-op `clearSuccessMessage()`.
- Fixed the stale `UserFormViewModel` comment referencing the removed `successMessage = 'Roles updated'`.

## Doc
- `command-feedback.md`: focus-to-banner note flipped to **exercised**; adoption status → **Phase 3 (in progress)** with the N4/N2 summary and remaining targets.

## Verification
- `tsc` + `eslint` (changed files) + `npm run build` + `npm run docs:check` green.
- 70 unit tests pass: users VM/service/banner suites (58) + `sanitizeCommandError` (8) + `CommandFeedbackEcho` (4).
[...]

PR #91, opened 2026-07-21
feat(command-feedback): Phase 3 PR2 — extract focus hook + migrate 4 sibling manage pages

lars-tice:
## Context

Second PR of the **command-feedback standard Phase 3 rollout** (`documentation/frontend/patterns/command-feedback.md`). Extracts the focus contract into a reusable hook and applies the standard uniformly to the four sibling manage pages. Builds on PR #90 (N4+N2).

## Extract `useCommandFeedbackFocus`
The reference page's focus contract (capture submit-trigger → focus banner → restore-on-dismiss, with disarm-on-cleanup) is pulled into `src/hooks/useCommandFeedbackFocus.ts` (stable identity via `useMemo`) + **5 unit tests**. `UsersManagePage` is refactored onto **two instances** (form banner + rich role-violation banner), behavior-preserving — removes ~50 lines of inline focus refs/effects.

## Migrate the 4 sibling manage pages
`Roles`, `Organizations`, `Org-Units`, `Schedules` — each now:
- `role="status"` success banner + `role="alert"` failure banner **guarded `!submissionError`** so exactly one alert region mounts (**INV-1**) + `aria-hidden` `CommandFeedbackEcho`.
- All mutations (create/update/deactivate/reactivate/delete; Organizations also contact/address/phone sub-entities) routed through `showCommandSuccess`/`showCommandFailure`.
- Previously-**raw** `submissionError` form banners now sanitized via `sanitizeCommandError` + focus-managed via `useCommandFeedbackFocus`.
- Pre-existing structured-error dialog branches (`HAS_CHILDREN`/`HAS_ROLES`/`HAS_USERS`/`STILL_ACTIVE`) preserved untouched.

**Folded in**: dropped the `OrganizationFormViewModel` bootstrap `toast.error` (it already set `submissionError` → Sonner-polite + banner = INV-1 double-surface) + its now-unused `sonner` import.

## Known follow-ups (out of scope, noted in the doc)
- Tier-2 surfaces (settings client-fields, assignments caseload, client intake, admin orphaned-deletions, auth reset-password) → PR 3.
- Organizations **create** flow (`OrganizationCreateForm` — a separate component/VM from the manage page).

## Verification
- `tsc` + `eslint` (all changed files) + `npm run build` green.
- **34 command-feedback unit tests pass** (hook 5, users VM/banner 17, sanitizer 8, echo 4).
- The 13 failing `useFocusAdvancement`/`useKeyboardNavigation` tests are **pre-existing** and unrelated — they reproduce identically with these changes stashed (verified).


frontend/src/pages/schedules/SchedulesManagePage.tsx
```diff
@@ -214,26 +244,45 @@ export const SchedulesManagePage: React.FC = observer(() => {
...
           await viewModel.refresh();
           await selectAndLoadTemplate(result.templateId);
         } catch (err) {
           log.error('Failed to transition after save', err);
-          setOperationError('Template was saved but failed to reload. Please refresh the page.');
+          showCommandFailure('Template was saved but failed to reload. Please refresh the page.');
         }
-      } else if (!result.success) {
-        setOperationError(result.error || 'Failed to save template');
+      } else if (formViewModel.submissionError) {
```
lars-tice, line 267: **[LOW — intentional behavior change, verify in UAT]** This `else if (formViewModel.submissionError)` replaces the old `else if (!result.success) setOperationError(result.error || 'Failed to save template')`. The VM's `submit()` validation early-return (`return { success:false, error:'Please fix validation errors before submitting' }`) does **not** set `submissionError` — only `touchAllFields()`/`validateAll()` populate the inline field errors. So a submit-time *validation* failure now surfaces **only** via inline field errors, not a page/form banner.

This is actually correct: it matches the reference `UsersManagePage` template exactly and aligns with the standard's scope note ("command results only; inline field-level validation is a separate concern"). `ScheduleFormFields` renders each field error with `role="alert"`, and Roles/Org-Units use `aria-invalid`+`aria-describedby`, so the failure is still surfaced. Flagging only because it's a real UX delta vs. the old code (same change in `RolesManagePage.tsx:373` and `OrganizationsManagePage.tsx:446`; Org-Units already behaved this way). Worth confirming in UAT that a screen-reader user submitting an invalid form still gets adequate feedback.

lars-tice (commented):
## Review — Phase 3 PR2 (focus hook + 4 sibling migrations)

**Verdict: APPROVE.** Rigorous pass against the command-feedback standard. No INV-1/2/3 violations, the extracted hook is sound, and the `UsersManagePage` refactor is behavior-preserving. Two low-severity notes below (one intentional-but-worth-flagging behavior change; one pre-existing).

### INV-1 — verified clean on all 4 siblings
The page error banner (`!submissionError` guard) and the form banner (renders `role="alert"` only when `submissionError` is truthy) are strictly complementary, so exactly one `role="alert"` mounts. `CommandFeedbackBanner` returns `null` on empty `message` and only emits `role="alert"` for `kind="error"` (success is `role="status"`), and the success banner is additionally guarded `!operationError && !error && !submissionError`. Confirmed identical in Roles, Organizations, Org-Units, Schedules.

### Hook & Users refactor — verified
`useCommandFeedbackFocus` capture-arm-restore-disarm is correct; `useMemo([captureTrigger, restore])` over two `useCallback([])` + a stable ref gives genuine stable identity. `restore()` consume-and-clears; the effect cleanup disarms on any teardown. The two-instance `UsersManagePage` split is equivalent to the prior single shared-ref implementation given the two banners are mutually exclusive by the `!submissionError` mount guard. `sanitizeCommandError` always returns a non-empty `display`, so `active` (`!!submissionError`) and `message` never disagree — no dropped focus/announcement. All structured dialog branches (HAS_CHILDREN/HAS_ROLES/HAS_USERS/STILL_ACTIVE) preserved; all command-failure surfaces sanitized; exhaustive-deps correct on every touched `useCallback`.

frontend/src/pages/organizations/OrganizationsManagePage.tsx
```diff
@@ -728,34 +780,31 @@ export const OrganizationsManagePage: React.FC = observer(() => {
...
-        {(listVM.error || operationError) && (
-          <div
-            className="mb-6 p-4 rounded-lg border border-red-300 bg-red-50"
-            role="alert"
+        {/* Error Banner — yields to the form-submit banner so exactly one
+            role="alert" region is ever mounted (INV-1). */}
+        {!formVM?.submissionError && (
+          <CommandFeedbackBanner
+            kind="error"
+            message={listVM.error || operationError}
```
lars-tice: **[LOW — pre-existing, not introduced here]** This page `CommandFeedbackBanner` renders `listVM.error || operationError` **unsanitized**. Command *mutation* failures now reach `operationError` via `reportFailure` (sanitized), but background *load* errors — `listVM.error`, and `operationError` set from the raw `newFormVM.submissionError` in `selectAndLoadOrg` (`:316`) — bypass `sanitizeCommandError`. Same shape on all four pages and the reference Users page, so out of scope here, but it's the one remaining path where a handler-internal string could reach the UI. Consider a follow-up to sanitize load-path errors too.

frontend/src/hooks/useCommandFeedbackFocus.ts
```diff
@@ -0,0 +1,72 @@
...
+    const el = restoreTargetRef.current;
+    restoreTargetRef.current = null;
+    if (el && document.contains(el)) {
+      el.focus();
+    }
+  }, []);
+
+  useEffect(() => {
+    if (active) {
+      restoreTargetRef.current = triggerCandidateRef.current;
```
lars-tice, line 61: Verified sound. `active:false->true` promotes the captured candidate to the armed restore target and focuses the banner; the cleanup disarms on any teardown (dismiss, unmount, or `active->false` driven elsewhere), so a background-load error banner that never activates can't arm a stale restore. `restore()` consume-and-clears (second call is a no-op). The 5 unit tests cover exactly these paths. One benign observation: `captureTrigger()` leaves `triggerCandidateRef` set after a *successful* submit, but since it's only promoted inside the `active` effect and every submit re-captures first, no stale trigger can ever be armed. No change requested.

frontend/src/pages/users/UsersManagePage.tsx
```diff
@@ -152,57 +153,20 @@ export const UsersManagePage: React.FC = observer(() => {
...
-  useEffect(() => {
-    if (richBannerActive) {
-      restoreTargetRef.current = submitTriggerCandidateRef.current;
-      usersErrorBannerRef.current?.focus();
-      return () => {
-        restoreTargetRef.current = null;
-      };
-    }
-  }, [richBannerActive]);
+  const richFocus = useCommandFeedbackFocus(richBannerActive);
```
lars-tice, line 169: Behavior-preservation confirmed. The old single shared `restoreTargetRef`/`submitTriggerCandidateRef` + two inline effects are replaced by two independent hook instances. Both call `captureTrigger()` at submit (same `document.activeElement`), but the form banner (`!!submissionError`) and the rich banner (`lastRoleViolations/lastRolePartialFailure`) are mutually exclusive — the page-level `UsersErrorBanner` is mount-guarded `!submissionError`, and role-violation submits null out `submissionError` — so only one instance ever arms/restores. Equivalent to the prior shared-ref behavior; the only divergence (both instances armed simultaneously) is unreachable given that guard.

lars-tice:
Architect review addressed (`7619f717`, APPROVE):

- **Finding #2** (LOW, pre-existing): sanitized the Organizations org-load-failure `setOperationError(newFormVM.submissionError …)` seed via `sanitizeCommandError(...)`. The broader raw-`viewModel.error` load-path leak (identical on the reference Users page + all siblings) is cross-cutting and pre-existing → seeded a focused uniform follow-up (`dev/active/seed-sanitize-load-path-errors-command-banner.md`) rather than expand this PR inconsistently.
- **Finding #1** (validation failures no longer show a page banner): no change — correct-by-design, consistent with the merged reference (#90) and the standard's field-validation scope; inline field errors still render. Will confirm in UAT.

`tsc` + `eslint` re-verified green. Ready to merge.

PR #92, opened 2026-07-22
feat(command-feedback): Phase 3 PR3 — Tier-2 surfaces (partial)

lars-tice:
## Context

Third PR of the **command-feedback standard Phase 3 rollout** (`documentation/frontend/patterns/command-feedback.md`). Covers the heterogeneous **Tier-2** surfaces. Unlike the uniform manage pages, these each got a per-surface treatment rather than a mechanical copy.

## Surfaces migrated (4)
- **`OrphanedDeletionsPage`** (admin) — was a raw error string with **no `role`** and **no success feedback** → sanitized `CommandFeedbackBanner` (error, `role="alert"`) + `role="status"` success ("Deletion workflow retry queued.") + `aria-hidden` echo on retry failure. **Also fixed a latent bug**: the retry-failure error was set then immediately wiped by the post-retry `fetchDeletions` (which clears `error` on start) — reordered so the outcome is applied *after* the refresh.
- **`UserCaseloadPage`** (assignments) — assign/unassign had **no feedback at all** → success banner + generic op-specific failure fallback + echo (the VM swallows the raw error and returns only a boolean; raw is logged in the VM). Load error (`vm.error`) sanitized and merged into one banner with `operationError` taking precedence (INV-1).
- **`ClientIntakePage`** — sanitized the failure surfaces (`vm.submitError` + per-item `vm.subEntityErrors`). Success navigates to the client detail page, so no same-page success banner.
- **`ResetPasswordPage`** — `submitError` sanitized; removed the `aria-live="polite"` that was downgrading the assertive `role="alert"`. **Kept `toast.success` as the documented navigation-away exception** (redirects to `/login` immediately, so a persistent banner would never be seen). `validationError` left as field-validation (out of scope).

## Deferred (seeded follow-ups, not defects)
- **Settings client-fields tabs** (`CategoriesTab`/`CustomFieldsTab`) — their `role="alert"` errors are **contextual** (right beside each form/row = good UX); the real gap is missing *success* feedback, which needs a surface *design* in a shared-VM tabbed page. Deferred to `dev/active/seed-command-feedback-settings-client-fields-tabs.md` (explicitly: do **not** consolidate the contextual errors).
- **Organizations *create* flow** (`OrganizationCreateForm`) and **Users load-path** (`UsersErrorBanner`) — pre-existing seeded follow-ups.

## Verification
- `tsc` + `eslint` (all changed files) + `npm run build` green.
- Command-feedback unit tests pass (hook 5, sanitizer 8, echo 4).
- Pre-existing unrelated `useFocusAdvancement`/`useKeyboardNavigation` failures remain (not touched).


frontend/src/pages/admin/OrphanedDeletionsPage.tsx
```diff
@@ -149,16 +167,26 @@ export const OrphanedDeletionsPage: React.FC = () => {
...
-        <div className="bg-red-50 border border-red-200 rounded-lg p-4 flex items-start gap-3">
-          <AlertTriangle size={20} className="text-red-500 mt-0.5 shrink-0" />
-          <div>
-            <p className="text-sm font-medium text-red-800">Error</p>
-            <p className="text-sm text-red-600">{error}</p>
-          </div>
-        </div>
+      {/* Command feedback — error takes precedence so exactly one live region
+          announces (INV-1); the echo is an aria-hidden visual copy. */}
+      {!successMessage && (
```
lars-tice: **Edge-case defect (LOW-MED): the retry command reuses the same `error` slot as the load path, and the 60s auto-refresh mutates it independently - two reachable holes.**

Unlike `UserCaseloadPage` (which correctly keeps a *separate* `operationError` distinct from `vm.error`), here the retry outcome is written into the same `error` state that `fetchDeletions()` clears on every load (line 50). The in-handler reorder fixes the *immediate* wipe, but the standalone `setInterval(fetchDeletions, 60_000)` (line 71) is not covered by it:

1. **Orphaned, undismissable echo.** Retry fails -> `error` banner + `aria-hidden` echo shown. 60s later the auto-refresh runs `setError(null)` -> the **banner vanishes but `echoMessage` is untouched** (nothing clears the echo on a load). The echo has no dismiss control of its own and `clearEcho()` is only reachable via the (now-gone) banner's `onDismiss` or the next retry - so a stale red toast is stuck on screen. This is exactly the banner/echo *desync* the standard says must never happen ("dismissing the banner clears the ViewModel error state **and** clears the echo ... so the two surfaces never desync").
2. **Zero feedback.** A persistent success banner is up (`successMessage` set, no auto-dismiss) -> an auto-refresh load then fails -> `error` becomes truthy while `successMessage` is still truthy. Both guards flip off (`{!successMessage && <error>}` and `{!error && <success>}`) and **neither banner renders** - the load failure is silent. (INV-1 is not violated - zero announce, not two - but it is a real visibility gap.)

**Fix options:** mirror `UserCaseloadPage` and give the retry command its own state slot (leave `error` for the load path only), or have `fetchDeletions` not clobber a live command error and clear the echo alongside `error`. Either removes both holes.

frontend/src/pages/admin/OrphanedDeletionsPage.tsx
```diff
@@ -62,21 +74,27 @@ export const OrphanedDeletionsPage: React.FC = () => {
...
+    // Refresh the list first (it clears `error`), THEN apply the retry outcome so
+    // a failure isn't immediately wiped by the refresh. The org id is a caller
+    // reference only — never interpolated into the displayed (sanitized) message.
     await fetchDeletions();
+    if (result.success) {
+      setError(null);
+      clearEcho();
+      setSuccessMessage('Deletion workflow retry queued.');
+    } else {
+      setError(reportFailure(result.error, { fallback: 'Retry failed. Please try again.' }));
```
lars-tice: The reorder itself is **correct** - applying the retry outcome *after* `await fetchDeletions()` reliably survives the refresh's `setError(null)`, and `reportFailure(result.error, ...)` sanitizes + `log.warn`s + drives the echo per the standard. The success branch's `setError(null)` also nicely suppresses any load error that surfaced during the intervening refresh. Retry is a per-row action (non-blocking), so *not* stealing focus is the right call per the decision tree. Only caveat is the separate auto-refresh path (see the banner-block comment).

frontend/src/pages/assignments/UserCaseloadPage.tsx
```diff
@@ -59,6 +74,14 @@ export const UserCaseloadPage: React.FC = observer(() => {
       setNewNotes('');
       setNewAssignedUntil('');
       setAssignReason('');
+      setOperationError(null);
+      clearEcho();
+      setSuccessMessage('Client assigned.');
+    } else {
+      setSuccessMessage(null);
+      setOperationError(
+        reportFailure(null, { fallback: 'Failed to assign client. Please try again.' })
```
lars-tice, line 84: Merging `operationError` (precedence) over the sanitized `vm.error` into one banner is sound for INV-1, and keeping `operationError` separate from the load-error `vm.error` is the right shape (the empty-state `!vm.isLoading && !vm.error` still works since `vm.error` is retained). Two low nits:

1. `reportFailure(null, {fallback})` - since the VM swallows the raw and returns only a boolean, passing `null` makes the hook emit `log.warn('assignments: command failed', { raw: '' })`: an empty, correlation-id-less log line that duplicates the *real* `log.error` the VM already wrote in its `catch`. Harmless but noisy/misleading. Given the VM is the only place with the raw, either accept the generic fallback without the redundant hook log, or (better, follow-up) have `assignClient`/`unassignClient` surface the error so this call site can pass it through.
2. The **assign form** failure is arguably form-blocking (the form stays open on failure), so per the standard's decision tree it would warrant focus-to-banner via `useEffect`. Acceptable to skip in this progressive pass (the `role="alert"` still announces in place, and unassign/retry are correctly non-blocking), but worth a note.

frontend/src/pages/clients/ClientIntakePage.tsx
```diff
@@ -212,7 +215,14 @@ export const ClientIntakePage: React.FC = observer(() => {
           role="alert"
           data-testid="intake-submit-error"
         >
-          <p className="text-sm text-red-700">{vm.submitError}</p>
+          <p className="text-sm text-red-700">
+            {
+              sanitizeCommandError(
```
lars-tice, line 240: Sanitization is correct and the navigation-away-success rationale is a defensible reading of the standard. Two observations (neither blocking):

1. **No `aria-hidden` echo and no focus-to-banner** were added, yet `intake-submit-error` (L212) is a form-blocking submit failure on a long, multi-section scrolling form - the exact case the echo (scroll-independence) and focus-move exist for. Fine to defer in this partial pass, but this is the surface that most wants them.
2. **Two `role="alert"` regions co-exist here** (`intake-sub-entity-warnings` at L192 + `intake-submit-error` at L212). If the VM can ever set `subEntityErrors` and `submitError` simultaneously that is a double-announce (INV-1). They look semantically mutually exclusive (registered-with-warnings vs. registration-failed) - please confirm the VM guarantees that. Pre-existing structure, not introduced here, but relevant to INV-1.

frontend/src/pages/auth/ResetPasswordPage.tsx
```diff
@@ -145,8 +146,9 @@ export const ResetPasswordPage: React.FC = () => {
       navigate('/login', { replace: true });
     } catch (err) {
       log.error('[ResetPasswordPage] Password update failed', err);
+      // Sanitize before display; the raw error is already logged above.
       setSubmitError(
-        err instanceof Error ? err.message : 'Failed to update password. Please try again.'
+        sanitizeCommandError(err, 'Failed to update password. Please try again.').display
```
lars-tice, line 151: **Correct on both counts.** (a) This sanitize + the retained raw `log.error` above satisfy the sanitization contract. (b) Removing `aria-live="polite"` from the error `role="alert"` (L255, this hunk) restores the assertive announcement - `role="alert"` carries implicit `aria-live="assertive"` + `aria-atomic="true"`, and the explicit `polite` was downgrading it. Keeping `toast.success` before the immediate `navigate('/login')` is the sanctioned navigation-away exception (a persistent banner would never be seen). Clean surface.

lars-tice (commented):
Phase-3 Tier-2 review (command-feedback standard). Overall solid; the per-surface judgment calls are mostly sound and the OrphanedDeletions reorder correctly fixes the described retry-error-wiped-by-refresh bug. Verdict: **APPROVE-WITH-FIXES** - one real edge-case defect on `OrphanedDeletionsPage` (the 60s auto-refresh shares the `error` slot with the retry command and can orphan the echo), plus a couple of low nits. INV-1 (no double-announce) holds on every surface I checked; ResetPassword's assertive restore is correct. Details inline.

lars-tice:
Architect review addressed (`867ccc11`, APPROVE-WITH-FIXES):

- **#1 (LOW-MED, real) — fixed**: OrphanedDeletions now gives the retry command its own `operationError` slot (never touched by `fetchDeletions`), paired with the echo; the load `error` stays owned by the fetch path. Banner shows `operationError ?? error` (command precedence), and a load failure clears a stale success banner so `error` XOR `success` always holds. This closes both holes — the auto-refresh can no longer orphan the echo (echo is set/cleared only with `operationError`) nor blank both banners.
- **#2 (LOW) — accepted per VM contract**: the empty `reportFailure(null,…)` `log.warn` on UserCaseload is because `AssignmentListViewModel.assignClient/unassignClient` swallow the raw error and return a boolean. The ideal fix (VM surfaces the error, giving users the real reason + a non-empty log) is a follow-up; the raw is already `log.error`'d in the VM.
- **#3 (LOW) — noted**: ClientIntake echo/focus-to-banner skipped for this partial pass (doc already marks ClientIntake partial). Confirmed the INV-1 question: `ClientIntakeFormViewModel` resets both `submitError` and `subEntityErrors` at submit start (`:595-597`), then sets `submitError` only on hard-failure/catch (`:612`/`:738`) and `subEntityErrors` only on post-creation partial success (`:722`) — disjoint branches, so at most one `role="alert"` mounts.

PR #93, opened 2026-07-22
feat(command-feedback): settings client-fields tabs (success + sanitize)

lars-tice:
## Context

Completes the last **Tier-2** command-feedback follow-up (seed: `dev/active/seed-command-feedback-settings-client-fields-tabs.md`) — the `CategoriesTab` / `CustomFieldsTab` under client-field settings. Their gap was **missing success feedback** and **unsanitized errors**. Their `role="alert"` errors are **contextual** (beside each form/row) and are good UX, so this is *additive* — not a consolidation.

Investigation settled the one design question: the page already has a `role="status"` banner (`ClientFieldSettingsPage.tsx:293`) but it's at the **page bottom**, wired to a *different* tab's batch "Save Changes" flow — so reusing it would be distant and semantically wrong. Per-item CRUD success gets its own **top-of-tabpanel** banner.

## Changes
- **VM** (`ClientFieldSettingsViewModel`): new `successMessage` observable (distinct from the batch-save `saveSuccess`); each category/field CRUD success sets a specific message; nulled at every mutation start, on `setActiveTab`, and via `clearSuccessMessage()` — **supersede, no timer** (INV-3).
- **Tabs**: top-of-tabpanel `<CommandFeedbackBanner kind="success">` guarded on `!<its three errors>` so an assertive error always wins (**INV-1**). The six contextual error displays wrapped in `sanitizeCommandError(...).display` (display-layer per the standard; VM keeps raw for logs). **Contextual placement unchanged.**
- **Bonus**: `deactivateCategory` / `deactivateCustomField` were silent on **both** success and failure — brought in line with the other lifecycle methods (set the lifecycle error on failure, success message on success).
- **Tests**: +3 (success set / superseded-by-error / cleared-on-tab-switch). Also completed the latent-incomplete `createMockService` (6 missing `IClientFieldService` methods) so the test file typechecks under the pre-commit hook (the app tsconfig excludes tests, so this was never caught).

## Verification
- `tsc` + `eslint` + `npm run build` + `docs:check` green.
- **67 VM tests pass** (64 existing + 3 new).

## Remaining follow-ups (after this)
Only two low-priority items left in the epic: the Organizations *create* flow (`OrganizationCreateForm`) and the Users load-path sanitization (`UsersErrorBanner`).


frontend/src/pages/settings/client-fields/CategoriesTab.tsx
```diff
@@ -268,14 +270,32 @@ export const CategoriesTab: React.FC<CategoriesTabProps> = observer(
               ))}
             </div>
 
+            {/* Success (polite) — yields to any contextual error so an assertive
+                failure always wins (INV-1). */}
+            {!viewModel.categoryLifecycleError &&
+              !viewModel.createCategoryError &&
+              !viewModel.updateCategoryError && (
```
lars-tice, line 277: **Finding 1 (LOW-MED) — a lingering sibling error masks a fresh success.**

The guard strictly satisfies INV-1 (at most one region renders). But it applies "assertive failure supersedes concurrent success" to a *stale* failure, not a concurrent one. Trace:

1. `deactivateCategory` fails → `categoryLifecycleError` set, contextual `role=alert` shows. It has **no dismiss control**, so it persists until the next *lifecycle* op (or a reactivate/delete click calls `clearCategoryLifecycleError()`).
2. User instead does a `createCategory` that **succeeds**. `createCategory` clears only `createCategoryError` + sets `successMessage` — it does **not** clear `categoryLifecycleError`.
3. Render: `!categoryLifecycleError` is false → success banner suppressed; the stale lifecycle `role=alert` still shows.

Net: a successful create shows **no** success confirmation and a **stale** error as if current. Same class for `updateCategory` succeeding while `createCategoryError` lingers (and mirror on CustomFieldsTab).

Per INV-3 ("stays visible until the user dismisses it **or issues the next command on that surface**"), the fix belongs in the VM: at each mutation start clear **all** of that tab's error observables, not just the method's own. e.g. `createCategory`/`updateCategory` should also null `categoryLifecycleError`; the lifecycle methods should null `create/updateCategoryError`. That makes the guard reflect only genuinely-concurrent state.

frontend/src/pages/settings/client-fields/CustomFieldsTab.tsx
```diff
@@ -415,7 +435,10 @@ export const CustomFieldsTab: React.FC<CustomFieldsTabProps> = observer(
                     data-testid="cf-error-alert"
                   >
                     <AlertCircle size={14} />
-                    {viewModel.createFieldError}
+                    {
+                      sanitizeCommandError(viewModel.createFieldError, 'Failed to create field')
```
lars-tice, line 439: **Finding 2 (LOW) — sanitizer false-positive on user-interpolated names.**

Good call confirming the sanitizer is a no-op on friendly text in the common case. But `createFieldError` is built in the VM by interpolating user input (`"${params.display_name}" already exists. Choose another name.`), and one of `INTERNAL_MARKERS` is `/\b[A-Z]\d{4}\b/` (SQLSTATE-shaped). A user-chosen name like `A1234` or `Room B2000` matches, so the sanitizer discards the specific, actionable guidance and shows the generic `'Failed to create field'` fallback. Verified: `"A1234"`, `"Room B2000"`, `"Ward C1000"` all match. Not a leak (fail-safe direction), just a legitimate message lost. Options: tighten the SQLSTATE marker (e.g. require a known class prefix / anchor to non-quoted context), or have the VM avoid re-embedding raw user input into a string the display regex then scans. Low severity — note, not a blocker.

frontend/src/viewModels/settings/__tests__/ClientFieldSettingsViewModel.test.ts
```diff
@@ -937,4 +943,46 @@ describe('ClientFieldSettingsViewModel', () => {
...
+    it('is superseded (nulled) by a subsequent failing lifecycle op', async () => {
+      await vm.createCategory('Custom Cat', 'custom_cat', 'org-123');
+      expect(vm.successMessage).toBe('Category created');
+
+      const failService = createMockService({
+        deactivateFieldCategory: vi.fn().mockResolvedValue({ success: false, error: 'In use' }),
+      });
+      vm = new ClientFieldSettingsViewModel(failService);
+      await vm.loadData('org-123');
+      vm.successMessage = 'Category created'; // simulate a lingering success
```
lars-tice: **Finding 3 (LOW) — coverage gap + test hygiene.**

The three new tests exercise set / superseded-by-failing-op / cleared-on-tab-switch — good. But the supersede test only covers the *failing op nulls success* direction. The reachable-but-unhandled case (Finding 1) — a **lingering error + a subsequent successful op**, where the success is masked — is not asserted. Adding that test would have surfaced Finding 1 and would lock the fix: set `categoryLifecycleError`, run a succeeding `createCategory`, assert the desired post-state (error cleared, `successMessage` set).

Minor: `vm.successMessage = 'Category created'` writes an observable outside an action. Harmless here, but prefer a setter / `runInAction` to avoid a strict-mode warning and match the VM's own discipline.

lars-tice (commented):
Review — command-feedback settings client-fields tabs (success + sanitize).

**Verdict: APPROVE-WITH-FIXES.** Core mechanic is sound: `successMessage` is nulled at every mutation start + `setActiveTab` + `clearSuccessMessage()` (INV-3, no timer), each failing lifecycle op nulls `successMessage` before setting its error, and the top-of-tabpanel success banner is guarded on the tab's three error observables so at most one live region renders (INV-1 holds strictly). The `deactivate*` restructure is correct and does not double-report with the `ConfirmDialog` (confirm handler closes the dialog unconditionally; the failure surfaces in the contextual `role=alert`). Not touching `isCategoryLifecycleActionInProgress` is fine — those confirm flows use local `isDeactivating` React state, unlike reactivate/delete which drive the dialog `isLoading` off the VM flag. 67 VM tests green.

Two real (low-severity) findings + one test-coverage gap inline. None block; author's own PR so COMMENT event.

**Ranked:**
1. (LOW-MED) A *lingering* sibling error masks a *fresh* success — a mutation clears only its own error, so an undismissed `categoryLifecycleError`/`createCategoryError` suppresses the success banner of an unrelated subsequent successful op. Per INV-3 the next command on the surface should clear the prior failure.
2. (LOW) Display-layer SQLSTATE regex `\b[A-Z]\d{4}\b` false-positives on user-interpolated names (e.g. a field named `A1234` / `Room B2000`), nuking the VM's friendly `createFieldError` guidance to the generic fallback.
3. (LOW) Test gap + hygiene: the supersede test only covers failing-op-nulls-success, not the inverse (lingering error + fresh success); and it writes `vm.successMessage` outside an action.

lars-tice:
Architect review addressed (`5d531d66`, APPROVE-WITH-FIXES):

- **#1 (LOW-MED) — fixed**: every category/field mutation now nulls **all three** of its tab's error observables at start (not just its own), so a lingering sibling error (e.g. a failed `deactivate` whose contextual error has no dismiss) can no longer mask a subsequent success. INV-3 supersede now holds in both directions.
- **#3 — fixed**: reworked the supersede test to use a real success (removed the out-of-action `vm.successMessage =` write) and **added the inverse test** (lingering error cleared by a fresh success — the exact Finding-1 case). 68 VM tests green.
- **#2 (LOW, fail-safe) — accepted/known**: the shared `sanitizeCommandError` SQLSTATE heuristic (`\b[A-Z]\d{4}\b`) can false-match a field name like `A1234` and drop the friendly "already exists, choose another name" guidance for the generic fallback. No leak (fail-safe). The fix belongs in the shared util + its 8 tests, so it's out of scope for this PR — flagging for a small follow-up on `sanitizeCommandError`.

PR #95, opened 2026-07-22
feat(command-feedback): migrate Organizations create flow (OrganizationCreateForm)

lars-tice:
## Context

The **last command-feedback surface**. `OrganizationCreateForm` (the org *create* flow — a separate component/VM from the manage page migrated in PR #91) still had a hand-rolled raw `submissionError` `role="alert"` banner.

## Change
- Raw `submissionError` banner → shared `<CommandFeedbackBanner kind="error">` (sanitized via `sanitizeCommandError`) + `aria-hidden` `CommandFeedbackEcho` + **focus-to-banner** via `useCommandFeedbackFocus`.
- `handleSubmit` captures the trigger, fires the echo (and `log.warn`s the raw) on failure, clears the echo on success.
- **Success navigates to the bootstrap status page** (like `ClientIntakePage`), so there's no same-page success banner.
- **No correlation_id** is surfaced — the bootstrap workflow client generates it per-request internally (unlike the read paths from PR #94).
- The `org-create-submit-error` testid is preserved.

## Verification
- `tsc` + `eslint` + `build` + `docs:check` green.

> ⚠️ **Pre-existing, unrelated:** `npm run test` shows 30 failing tests in `InvitationAcceptanceViewModel` (26) + `OrganizationFormViewModel` (4). These are **on `main`** — they reproduce identically with this change stashed, and neither imports `OrganizationCreateForm`. Not introduced here; flagged separately.

## Epic
This completes the command-feedback rollout across every surface. One tiny follow-up remains: the shared `sanitizeCommandError` SQLSTATE-heuristic false-positive.


frontend/src/pages/organizations/OrganizationCreateForm.tsx
```diff
@@ -199,6 +204,16 @@ export const OrganizationCreateForm: React.FC<OrganizationCreateFormProps> = obs
...
 
+    // Command-result feedback: sanitize + log + drive the aria-hidden echo, and
+    // move focus to the banner on a form-blocking submit failure. Success
+    // navigates to the bootstrap status page, so there's no same-page success.
+    const {
+      failed: reportFailure,
+      clear: clearEcho,
+      echoMessage,
+    } = useCommandFeedback('organization-create');
+    const submitFocus = useCommandFeedbackFocus(!!viewModel.submissionError);
```
lars-tice, line 219: **Confirmed correct (not a change request).** Keying `active` on `!!viewModel.submissionError` satisfies the standard's strictest focus rule: focus must key on an observable set *only* by the submit path. In this VM `submissionError` is set exclusively in `submit()` (and cleared) — a create form has no load path that could set it — so a background load can never steal focus to a stale trigger. Good.

frontend/src/pages/organizations/OrganizationCreateForm.tsx
```diff
@@ -223,13 +238,22 @@ export const OrganizationCreateForm: React.FC<OrganizationCreateFormProps> = obs
       }
 
       setIsSubmitting(true);
+      // Remember the trigger so a banner dismiss can restore focus to it.
+      submitFocus.captureTrigger();
```
lars-tice, line 246: `captureTrigger()` placement is right: it runs synchronously in the handler before `await viewModel.submit()`, while the submit button still holds focus and before the `setIsSubmitting(true)` re-render disables it. The captured ref survives the disable/re-enable cycle (button is re-enabled in `finally`), so `restore()` → `el.focus()` lands on it after dismiss.

frontend/src/pages/organizations/OrganizationCreateForm.tsx
```diff
@@ -266,34 +290,28 @@ export const OrganizationCreateForm: React.FC<OrganizationCreateFormProps> = obs
...
+              announcement); focus moves here on failure. Success navigates to the
+              bootstrap status page, so there's no same-page success banner. */}
+          <CommandFeedbackBanner
+            ref={submitFocus.bannerRef}
+            kind="error"
+            message={
+              viewModel.submissionError
+                ? sanitizeCommandError(
+                    viewModel.submissionError,
+                    'Failed to create organization. Please try again or contact support.'
```
lars-tice: **Nit (low, DRY):** the fallback string `'Failed to create organization. Please try again or contact support.'` is duplicated here (banner) and at L255 (`reportFailure`). They MUST stay identical for banner/echo text parity — which they currently are — but a future edit to one and not the other would silently desync the two surfaces. Consider hoisting to a `const SUBMIT_FALLBACK` at module scope and referencing it in both places.

frontend/src/pages/organizations/OrganizationCreateForm.tsx
```diff
@@ -266,34 +290,28 @@ export const OrganizationCreateForm: React.FC<OrganizationCreateFormProps> = obs
...
+                : null
+            }
+            onDismiss={() => {
+              viewModel.clearSubmissionError();
+              clearEcho();
+              submitFocus.restore();
+            }}
+            data-testid="org-create-submit-error"
+          />
+          <CommandFeedbackEcho message={echoMessage} />
```
lars-tice, line 317: Echo/banner text parity verified: banner = `sanitizeCommandError(submissionError, fallback).display`; echo = `reportFailure(submissionError, {fallback})` → `sanitizeCommandError(raw, fallback).display` internally. Same raw + same fallback + same pure sanitizer ⇒ identical display string. No desync.

frontend/src/pages/organizations/OrganizationCreateForm.tsx
```diff
@@ -266,34 +290,28 @@ export const OrganizationCreateForm: React.FC<OrganizationCreateFormProps> = obs
...
+                  ).display
+                : null
+            }
+            onDismiss={() => {
+              viewModel.clearSubmissionError();
+              clearEcho();
+              submitFocus.restore();
+            }}
+            data-testid="org-create-submit-error"
+          />
```
lars-tice, line 316: **Observation (minor, testing — not blocking):** this surface isn't covered by `e2e/command-feedback-a11y.spec.ts` (the `RUN_A11Y_GATE` spec references only the reference page). The DoD's "single announcement + zero `aria-hidden-focus` + focus-move/restore asserted" item is therefore unverified by CI for org-create. Consistent with how the other Phase-3 rollout pages shipped (progressive), so acceptable to defer — but worth a seed to extend the gate spec to cover the create form's banner focus/restore.

lars-tice (commented):
**Command-feedback rollout — final surface (`OrganizationCreateForm`). Verdict: APPROVE.**

Clean, faithful migration. All four invariants verified against the standard (`documentation/frontend/patterns/command-feedback.md`):

- **INV-1 (single announcement)** ✓ — the `CommandFeedbackBanner kind="error"` (`role="alert"`) is the only announcing region on this surface. `CommandFeedbackEcho` is `aria-hidden`; the validation-summary div (L604) carries no role; field-level `role="alert"`s (e.g. `SubdomainInput`) are field-scoped (`aria-invalid`+`aria-describedby`) and only appear on validation failure, so they never co-mount with the post-validation command banner. Out of scope per the standard.
- **Focus contract** ✓ — `active` keys on `!!viewModel.submissionError`, which in this VM is set *only* by `submit()` (no load path on a create form), exactly satisfying the standard's "key focus on a submit-only observable" rule. `captureTrigger()` fires before the await while the submit button still holds focus; focus moves via the hook's `useEffect`, no `setTimeout`; `restore()` on dismiss.
- **Failure detection** ✓ — validation-fail is short-circuited up-front (L233) and surfaces via `validationErrors` (summary + per-field), so the `submit()===null && !submissionError` tail is unreachable/defensive. The `else if (submissionError)` branch fires only on real workflow errors.
- **Sanitization + consistency** ✓ — banner renders `sanitizeCommandError(raw, fallback).display`; echo goes through `reportFailure` → same sanitizer + same raw + same fallback → identical string, and `log.warn`s the raw. No raw leak to the UI.

The **no-same-page-success-banner** (navigates to bootstrap status, like `ClientIntakePage`) and **no-correlation-id** (the workflow client mints it per-request internally; nothing to thread at this layer) are reasonable, documented scope calls — agreed.

Only nits below (non-blocking). The 30 pre-existing `OrganizationFormViewModel`/`InvitationAcceptanceViewModel` failures are on main and unrelated to this diff — not attributed here.

lars-tice:
Architect review addressed (`628528c9`, APPROVE):

- **#1 (LOW/DRY) — fixed**: hoisted the duplicated submit-error fallback string to a module-scope `SUBMIT_ERROR_FALLBACK` const, so the banner (`sanitizeCommandError`) and echo (`reportFailure`) can't desync on a one-sided edit.
- **#2 (Minor/testing) — seeded**: the F4 a11y gate spec only drives the reference page; extending it to all migrated surfaces is an epic-wide coverage gap (not #95-specific). Seeded `dev/active/seed-extend-command-feedback-a11y-gate.md`.

